PR #3, opened 2022-08-15
Implement provider-cloudscale package

ccremer:
* Adds `pkg.provider.cloudscale` with some defaults for `cloudscale-lpg-2` configs.

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/tutorials/install-cloudscale.adoc
+++ b/docs/modules/ROOT/pages/tutorials/install-cloudscale.adoc
@@ -51,7 +51,7 @@ vault kv patch "${parent}" token=<the-cloudscale-token>
 classes:
   - appcat.provider.cloudscale
 applications:
-  - pkg.provider.cloudscale <1>
+  - pkg.appcat.provider.cloudscale <1>
 
 parameters:
   packages:
--- a/docs/modules/ROOT/pages/tutorials/install-cloudscale.adoc
+++ b/docs/modules/ROOT/pages/tutorials/install-cloudscale.adoc
@@ -1,0 +1,68 @@
+= Install Cloudscale Stack
+
+This tutorial will guide you through the installation process of installing the following parts:
+
+- A Crossplane provider to manage cloudscale.ch resources.
+//- A Crossplane composite (XRD) which serves as the user's "interface" (API) for ordering S3 buckets.
+//- A Crossplane composition that "implements" an XRD using the cloudscale.ch provider.
+
+== Requirements
+
+- `commodore`
+- `vault`
+- A Syn-enabled cluster with ArgoCD
+- Vault instance to store credentials
+
+== Preparation
+
+. Prepare some variables
++
+[source,bash]
+----
+export CLUSTER_ID=c-appuio-cloudscale-lpg-2
+export TENANT_ID=t-falling-frog-2211
+export ZONE=cloudscale-lpg-2
+----
+
+. Login to Vault.
++
+[source,bash]
+----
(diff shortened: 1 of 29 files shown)


ccremer:
(PR currently stacks upon #4 )

packages/provider/cloudscale.yml
```diff
@@ -0,0 +1,36 @@
...
+              # This is not resolving when during "commodore package compile"
+              namespace: ${crossplane:namespace}
+
+  appcat:
+    secrets:
+      cloudscale-lpg-2:
+        metadata:
+          namespace: ${crossplane:namespace}
+        stringData:
+          CLOUDSCALE_API_TOKEN: '?{vaultkv:${cluster:tenant}/${cluster:name}/appcat/objectstorage/provider-cloudscale/cloudscale-lpg-2}'
```
ccremer, line 31: I went with explicitly defining our clusters here, with the intention that some point we will have cross-cluster and cross-cloud-provider configs, where users can deploy buckets in other cloud providers by switching out the composition.

packages/provider/common.yml
```diff
@@ -0,0 +1,14 @@
...
+
+parameters:
+  components:
+    crossplane:
+      url: https://github.com/projectsyn/component-crossplane.git
+      version: v2.2.0
+
+    appcat:
+      url: https://github.com/vshn/component-appcat.git
+      version: provider-cloudscale
```
ccremer, line 13: Unfortunately, we can't use something like `HEAD` to reference a current local state.

docs/modules/ROOT/pages/tutorials/install-cloudscale.adoc
```diff
@@ -0,0 +1,68 @@
+= Install Cloudscale Stack
+
+This tutorial will guide you through the installation process of installing the following parts:
+
+- A Crossplane provider to manage cloudscale.ch resources.
+//- A Crossplane composite (XRD) which serves as the user's "interface" (API) for ordering S3 buckets.
+//- A Crossplane composition that "implements" an XRD using the cloudscale.ch provider.
```
ccremer, line 7: Note: I haven't tested this tutorial yet. I will fix missing or wrong steps when we roll it out, in a different story.

packages/provider/common.yml
simu: Currently components are fetched from the provided URL by `package compile`, so local state can't be referenced.

packages/provider/cloudscale.yml
ccremer: On second thought, this is actually not required.

For example: If there's `cloudscale-rma-1` cluster, then the API token should point to a project like `appcat-objects-cloudscale-rma-1`. Also, since in the resources (at least for S3) users can still choose a different bucket region. Yet that bucket and objectsuser should remain in the `appcat-objects-cloudscale-rma-1` project, even if the bucket ends up in `lpg`.

For cross-provider-usage: Let's assume we have a cluster `exoscale-ch-gva-2` and we want to support creating buckets on cloudscale's `lpg` region (e.g. for offsite-backup buckets). This time we create a cloudscale Project named `appcat-objects-exoscale-ch-gva-2` and it contains the objects users and bucket of `exoscale-ch-gva-2`. There's no link to `cloudscale-lpg-2` in any way.

packages/provider/common.yml
ccremer: Yes, I'm aware and that's fine for now.
Let's deal with that problem separately.

Kidswiss (approved):
LGTM

Although not really proficient in Commodore Components :)

PR #4, opened 2022-08-15
Render additional resources in component

ccremer:
* Adds `secrets` parameter
* Adds `additionalResources` parameter
* Adds reference documentation

Needed so that the following to-be-implemented packages can be fully deployed with stuff that `component-crossplane` doesn't (yet) support. Later it might be worthwile to move those features, but it's good enough for now.

## Checklist


ZuGao:
What is the reason behind this component?

ZuGao (approved):
LGTM

PR #8, opened 2022-08-18
Add support for adding compositions in component

ccremer:
* Adds a feature where Crossplane compositions can be rendered using the component, by providing the full composition spec in `appcat.compositions`.
* Adds parameter reference docs

This PR is preparation for a package that adds a composition for cloudscale objectstorage in a follow-up

## Checklist

### DIFF
--- a/component/component/main.jsonnet
+++ b/component/component/main.jsonnet
@@ -17,16 +17,19 @@ local sync_options = {
 
 // https://syn.tools/syn/explanations/commodore-components/secrets.html
 local secrets = std.filter(function(it) it != null, [
-  if params.secrets[s] != null then
-    kube.Secret(s) {} + com.makeMergeable(params.secrets[s])
-  for s in std.objectFields(params.secrets)
+  if params.secrets[name] != null then
+    local secret = params.secrets[name];
+    assert std.objectHas(secret, 'metadata') : "missing `.metadata` in secret '%s'" % name;
+    assert std.get(secret.metadata, 'namespace', '') != '' : "`.metadata.namespace` in secret '%s' cannot be empty" % name;
+    kube.Secret(name) {} + com.makeMergeable(secret)
+  for name in std.objectFields(params.secrets)
 ]);
 
 local additionalResources = std.filter(function(it) it != null, [
-  if params.additionalResources[s] != null then
-    local res = params.additionalResources[s];
-    kube._Object(res.apiVersion, res.kind, s) + com.makeMergeable(res)
-  for s in std.objectFields(params.additionalResources)
+  if params.additionalResources[name] != null then
+    local res = params.additionalResources[name];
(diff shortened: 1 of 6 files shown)


component/component/main.jsonnet
```diff
@@ -16,29 +16,40 @@ local sync_options = {
...
     local res = params.composites[name];
     kube._Object('apiextensions.crossplane.io/v1', 'CompositeResourceDefinition', name) + sync_options + com.makeMergeable(res)
   for name in std.objectFields(params.composites)
-];
+]);
+
+local compositions = std.filter(function(it) it != null, [
+  if params.compositions[name] != null then
+    local composition = params.compositions[name];
+    kube._Object('apiextensions.crossplane.io/v1', 'Composition', name) + sync_options + com.makeMergeable(composition)
```
ZuGao, line 45: You could export this logic to an external library how it was done for component-crossplane. Take a look [here](https://github.com/projectsyn/component-crossplane).

component/tests/defaults.yml
```diff
@@ -29,3 +29,50 @@ parameters:
               schema:
                 openAPIV3Schema:
                   type: object
+
+    compositions:
+      test-composition:
```
ZuGao, line 34: Since we plan to have a lot of composistions here, does it make sense to group them by domain, like `s3-compositions`?

component/component/main.jsonnet
ccremer: As long as it's just one occurrence I don't see a need to put that into a function.
I also didn't want to introduce a hard dependency on component-crossplane's exported library just for 1 line of code :)

component/tests/defaults.yml
ccremer: I'm not sure what that grouping by domain would actually achieve, Where do you see the value?
I mean we could just prefix the name of compositions according to their domain.

component/component/main.jsonnet
ZuGao: We might use more complex functions later on for composition. For now it is ok to skip now.

component/tests/defaults.yml
ZuGao: It will be easier to figure out where the compositions belong to.

component/component/main.jsonnet
ccremer: Yes, but that's not relevant yet for this PR.

component/tests/defaults.yml
ccremer: Please have a look at the next PR that adds the cloudscale composition: https://github.com/vshn/component-appcat/blob/pkg-composition/packages/composition/objectstorage/cloudscale.yml#L8-L13
Is it not enough to figure where it belongs? It's in the spec which XRD is being implemented.

ccremer: I will go forward with the current implementation for now. We can still discuss this topic in the next stage.

PR #16, opened 2022-09-07
Propagate CRD status conditions to claim object

ZuGao:
This PR resolves the crossplane's issue that it does not pass the error conditions from the CRDs to the composite. An [issue](https://github.com/crossplane/crossplane/issues/3262) has been opened earlier to crossplane for that matter but we are not sure if the crossplane community will decide to resolve it the way we want.

## Checklist


tests/kuttl/patch-object.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -eo pipefail
+set -x
+
+
+sleep 3
```
ZuGao, line 7: I'm open for a suggestion here. Unfortunately I could not find a solution with `kubectl wait` command for situations when resources does not exist yet. 

Insight: The CRDs are being created later then this script execution which means I get empty names for the CRDs.

Kidswiss: You said, you also need some timeout for this thing?

Maybe something like this would work:
```
for (( i=1; i<=5; i++ )); do sleep 1 && kubectl get objectuser; done
```

This should try 5 seconds to get objectuser and then return.

ZuGao: Ok we went with the kuttle TestStep and used `sleep` :)

Kidswiss (approved):

tests/kuttl/exoscale-status-test/00-install-exoscale-composition.yaml
```diff
@@ -0,0 +1,313 @@
...
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: '10'
+  labels:
+    name: xobjectbuckets.appcat.vshn.io
+  name: xobjectbuckets.appcat.vshn.io
+spec:
+  claimNames:
+    kind: ObjectBucket
+    plural: objectbuckets
```
ccremer, line 13: This feels like a manually maintained duplication from the golden tests of the packages.
Any thoughts about automating/generating it from the golden tests?
I'd rather not maintain 2 versions of a composition, as the test would be worthless since it's not testing the result of a commodore package.

ccremer:
Also, this PR adds e2e tests as well as the status condition propagation. I'd like to see 2 separate PRs
- One that adds e2e tests in general (with the boilerplate, but without the condition propagation)
- This PR that focuses solely on condition propagation.

tests/kuttl/cloudscale-status-test/03-wait-crds.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  # Wait for crossplane create the CRDs
+  - command: sleep 3
```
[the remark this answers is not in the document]
ZuGao: I meant generated resources, do you have another solution for that? I tried with TestAssert but I think it does not work on cluster scoped resources.

tests/kuttl/exoscale-status-test/00-install-exoscale-composition.yaml
ZuGao: I thought about it and even started to work into that direction but I realised that those are 2 different kind of tests and they can and will deviate over time.  What I could do is to compile the package and copy the result but again I have to change namespaces or add fields such as annotations. I would really like to leave them separate or at worst to get them from compile and adjust the fields.

tests/kuttl/cloudscale-status-test/03-wait-crds.yaml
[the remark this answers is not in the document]
ccremer: > I meant generated resources

Please change the comment accordingly, `CRDs` is wrong.

> I tried with TestAssert but I think it does not work on cluster scoped resources.

That doesn't make sense. In provider-cloudscale we can test & assert cluster-scoped resources without issues: https://github.com/vshn/provider-cloudscale/blob/master/test/e2e/provider/01-assert.yaml

ccremer (changes requested):
(please see comments and conversations from previous reviews)

ZuGao:
I opened a new [PR](https://github.com/vshn/component-appcat/pull/17) to configure the integration tests. This one will be adapted as soon as the other one is reviewed and accepted.

tests/kuttl/exoscale-status-test/00-install-exoscale-composition.yaml
ccremer: > 2 different kind of tests and they can and will deviate over time

How so? What do you expect to change that is different from a package output vs the test case?

> to get them from compile and adjust the fields.

Then please do this, e.g. somewhere in a make target, using `yq` or so. But having them separate is a no-go IMO. We add the integration test for the package, because that's the thing that users change. It's too easy to forgot about adjusting the test case when users change configuration in a package (e.g. add a patch in composition) and that creates a false security.

ZuGao: Actually I could argue to remove everything from the spec object and test only the conditions, in that case we do not have to maintain neither the kuttl install file nor the kuttl assert file. With this approach we do not have to maintain the test at all, we will test the conditions that they are propagated accordingly.

tests/kuttl/cloudscale-status-test/03-wait-crds.yaml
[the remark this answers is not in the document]
ZuGao: I will try to do that, the problem with kuttl in this component is that it required time for crossplane to reconcile resources. Let's see maybe I can make it work with a timeout in TestAssert + static namespaces.

tests/kuttl/exoscale-status-test/00-install-exoscale-composition.yaml
ccremer: > Updated: In any case I think I need to generate the composition and composite to make sure any changes to the composite and composition are reflected in the install step.

Yes, that's what I meant the whole time :slightly_smiling_face:

ZuGao: Now they are autogenerated.

tests/crds/cloudscale.crossplane.io_objectsusers.yaml
```diff
@@ -0,0 +1,314 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.9.2
```
ccremer, line 6: Please put the `tests/crds` to `.gitignore`, there's no value to vendor them here.

tests/local.mk
```diff
@@ -8,6 +8,8 @@ registry_sentinel = $(kind_dir)/registry_sentinel
 cloudscale_version = $(shell yq -e '.parameters."pkg.appcat.provider.cloudscale".images.provider-cloudscale.tag' packages/provider/cloudscale.yml)
 exoscale_version = master
 
+golden_folder = packages/tests/golden
```
ccremer, line 11:
```suggestion
golden_dir = packages/tests/golden
```

tests/local.mk
```diff
@@ -36,6 +38,22 @@ install-crds: kind-setup # install cloudscale and exoscale CRDs
 	@kubectl apply -f "https://raw.githubusercontent.com/vshn/provider-exoscale/"${exoscale_version}"/package/crds/exoscale.crossplane.io_buckets.yaml"
 	@kubectl apply -f "https://raw.githubusercontent.com/vshn/provider-exoscale/"${exoscale_version}"/package/crds/exoscale.crossplane.io_iamkeys.yaml"
 
+.exoscale-composition:
+	$(MAKE)	.prepare-integration-tests -e instance=exoscale
+
+.cloudscale-composition:
+	$(MAKE)	.prepare-integration-tests -e instance=cloudscale
```
ccremer, line 45:
```suggestion
.exoscale-composition:
	$(MAKE) .prepare-integration-tests -e instance=exoscale

.cloudscale-composition:
	$(MAKE) .prepare-integration-tests -e instance=cloudscale
```

tests/local.mk
```diff
@@ -47,7 +65,7 @@ $(kuttl_bin): | $(go_bin)
 	go install github.com/kudobuilder/kuttl/cmd/kubectl-kuttl@latest
 
 test-integration: export KUBECONFIG = $(KIND_KUBECONFIG)
-test-integration: $(kuttl_bin) local-install ## Run integration tests with kuttl
+test-integration: $(kuttl_bin) local-install generate-integration-compositions
```
ccremer, line 68: Only if you add a comment with `##`, it will show up in `make help`
```suggestion
test-integration: $(kuttl_bin) local-install generate-integration-compositions ## Run integration tests with kuttl
```

tests/local.mk
```diff
@@ -36,6 +38,22 @@ install-crds: kind-setup # install cloudscale and exoscale CRDs
...
+	$(MAKE)	.prepare-integration-tests -e instance=exoscale
+
+.cloudscale-composition:
+	$(MAKE)	.prepare-integration-tests -e instance=cloudscale
+
+.PHONY: .prepare-integration-tests
+.prepare-integration-tests:
+	rm -rf .cache compiled dependencies vendor
+	cp $(golden_folder)/composite-objectstorage-$(instance)/appcat/appcat/composites.yaml tests/kuttl/$(instance)-status-test/00-install-$(instance)-composite.yaml
+	yq e '.spec.writeConnectionSecretsToNamespace = "default"' $(golden_folder)/composition-objectstorage-$(instance)/appcat/appcat/compositions.yaml > tests/kuttl/$(instance)-status-test/00-install-$(instance)-composition.yaml
```
ccremer, line 51:
```suggestion
	rm -rf .cache compiled dependencies vendor
	cp $(golden_dir)/composite-objectstorage-$(instance)/appcat/appcat/composites.yaml tests/kuttl/$(instance)-status-test/00-install-$(instance)-composite.yaml
	yq e '.spec.writeConnectionSecretsToNamespace = "default"' $(golden_dir)/composition-objectstorage-$(instance)/appcat/appcat/compositions.yaml > tests/kuttl/$(instance)-status-test/00-install-$(instance)-composition.yaml
```

tests/kuttl/patch-object.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -eo pipefail
+set -x
+
+user_type=${1}
+bucket_type=${2}
+
+user_name=$(kubectl get ${user_type} -oname)
+bucket_name=$(kubectl get ${bucket_type} -oname)
```
ccremer, line 10: I generally recommend to put a space between the option and the argument value for easier readability.
```suggestion
user_name=$(kubectl get ${user_type} -o name)
bucket_name=$(kubectl get ${bucket_type} -o name)
```

ccremer (changes requested):

ccremer (approved):

PR #19, opened 2022-09-12
Define the version used via hierarchy

ccremer:
This PR changes packages so that `component-appcat` in `parameters.components` needs be defined outside of the package (e.g. commodore defaults).

Long story why:

The issue is the version number combined with the monorepo (component & package in one repo). When the package is using a tagged version (say, `0.1.1`) that references `component-appcat` in `master` version, the `master` version is used for the component, even if the commodore defaults reference a different version. Not only are the packages used not on the same commit anymore with the component, but it also breaks the release process by introducing a chicken-and-egg-problem.

When releasing a new version, we would need to make a PR or commit that references `component-appcat` in the next version that doesn't yet exist (say, `0.1.0` > `0.2.0`). The CI checks for that PR will fail, but after merging we'll create the new tag (`0.2.0`).

Having the component-appcat version outside of the packages, breaks a bit the philosophy behind packages, as one does need to maintain not only the right package version (that should include/define all dependencies) but also the version for the component.

We can alleviate this a bit by passing through the package version used to a intermediate parameter before setting the version in `components`

## Checklist


ccremer:
For reviewers: The important change is in `packages/composition/objectstorage/common.yml` and maybe the documentation page.

bastjan (approved):

PR #29, opened 2022-10-14
Add endpoint to Bucket status

davidgubler:
Add new "endpoint" parameter which contains the host name of the endpoint. Repurpose the current "endpointURL" parameter to contain the endpoint URL including the scheme. This works in conjunction with https://github.com/vshn/provider-cloudscale/pull/51 and https://github.com/vshn/provider-exoscale/pull/42

### DIFF
--- a/packages/composition/objectstorage/cloudscale.yml
+++ b/packages/composition/objectstorage/cloudscale.yml
@@ -94,9 +94,12 @@ parameters:
                     region: ""
                     bucketDeletionPolicy: ${pkg.appcat.composition.objectstorage.cloudscale:bucketDeletionPolicy}
               connectionDetails:
-                - fromFieldPath: spec.forProvider.endpointURL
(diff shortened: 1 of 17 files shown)


packages/composition/objectstorage/cloudscale.yml
```diff
@@ -90,13 +90,17 @@ parameters:
                     credentialsSecretRef:
                       name: ""
                       namespace: ${pkg.appcat.composition.objectstorage.cloudscale:secretNamespaces:objectsusers}
+                    endpoint: ""
```
ccremer, line 93: This field doesn't exist
```suggestion
```

ccremer (approved):
One issue, rest lgtm

packages/composition/objectstorage/cloudscale.yml
davidgubler: Fixed, thanks.

PR #31, opened 2022-10-18
Improve testing experience with Kuttl

ccremer:
* Cleans up kuttl testing
* Suppresses events and debug logs. Kuttl doesn't redirect assertion errors to stderr while outputting the events to a file, so we rather not display the events.
* Adds a bit of README
* Increases default timeout of kuttl tests. 30s seems a bit low
* Forces Helm update when installing Crossplane to avoid a version mismatch between GH actions and local env.

## Checklist


davidgubler:
It definitely improves the output a lot, thanks!

However, it is still unreliable. When I switched to the branch and did "make test-integration", I got a diff and both tests failed. Then I did "make clean; helm repo update" (not sure if the latter did something; the output of "helm repo update" is useless because it prints "Successfully got an update from .... chart repository" no matter whether it actually updated something or not!) and successive runs of "make test-integration" then passed.

That shouldn't prevent this PR from being merged though because it is still a huge improvement.

PR #33, opened 2022-10-20
Create XRDs using Kubebuilder

ccremer:
Adds boilerplate so that the XRDs can be generated using Go and Kubebuilder.
This removes the need to manually craft the OpenAPI scheme for the XRDs, reducing human mistakes.

Closes: #15 

## Checklist

- [x] PR contains a single logical change (to build a better changelog).
- [x] Update the documentation.
- [x] Categorize the PR by setting a good title and adding one of the labels:
      `bug`, `enhancement`, `documentation`, `change`, `breaking`, `dependency`
      as they show up in the changelog.
- [x] Link this PR to related issues or PRs.

PR #32, opened 2022-10-19
Add PostgreSQL composition

ZuGao:
* Add composition for exoscale postgres database.
* Added commodore config parameters for the package
* Added golden test

## Checklist

- [x] PR contains a single logical change (to build a better changelog).
- [x] Update the documentation.
- [x] Categorize the PR by setting a good title and adding one of the labels:
      `bug`, `enhancement`, `documentation`, `change`, `breaking`, `dependency`
      as they show up in the changelog.

PR #36, opened 2022-10-26
Add ExoscalePostgreSQL composite

ZuGao:
* Added postgres compoiste
* Update golden tests
* Add parameter docs for the syn package

Update:
The mandatory fields are not clear yet. I set the PR reviewable as soon as I have that information.

Update 2:
Ready to be reviewed

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/references/composite/dbaas/exoscale/postgres.adoc
+++ b/docs/modules/ROOT/pages/references/composite/dbaas/exoscale/postgres.adoc
@@ -1,0 +1,3 @@
+= pkg.appcat.composite.dbaas.exoscale.postgres
+
+The parent key for all of the following parameters is`pkg.appcat.composite.dbaas.exoscale.postgres`.
--- a/packages/composite/dbaas/exoscale/postgres.yml
+++ b/packages/composite/dbaas/exoscale/postgres.yml
@@ -1,0 +1,173 @@
+classes:
+  - ...common
+
+parameters:
+
+  appcat:
+    composites:
+      "exoscalepostgresqls.appcat.vshn.io":
+        spec:
+          group: appcat.vshn.io
+          names:
+            kind: XExoscalePostgreSQL
+            plural: xexoscalepostgresqls
+          claimNames:
+            kind: ExoscalePostgreSQL
+            plural: exoscalepostgresqls
+          connectionSecretKeys:
+            - POSTGRESQL_URL
+            - POSTGRESQL_DB
+            - POSTGRESQL_HOST
+            - POSTGRESQL_PORT
+            - POSTGRESQL_USER
+            - POSTGRESQL_PASSWORD
+            - ca.crt
+          defaultCompositionRef:
+            name: exoscale.postgres.appcat.vshn.io
+          versions:
+            - name: v1
+              referenceable: true
+              served: true
+              schema:
+                openAPIV3Schema:
+                  type: object
+                  properties:
+                    spec:
+                      type: object
+                      properties:
+                        parameters:
+                          type: object
+                          properties:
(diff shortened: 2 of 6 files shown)


Kidswiss:
FWIW: with https://github.com/vshn/component-appcat/pull/33 it's possible to generate the XRD from a go struct.

But I'm fine with merging this as is and creating a tech debt to migrate later.

ZuGao:
Let's do a full migration in another PR. I worked on this composite before #33

packages/composite/dbaas/exoscale/postgres.yml
```diff
@@ -0,0 +1,176 @@
+classes:
+  - ...common
+
+parameters:
+
+  pkg.appcat.composite.exoscalepostgresql:
+    defaultCompositionRef: "${facts:cloud}.exoscalepostgresqls.appcat.vshn.io"
```
ccremer, line 7: This doesn't make sense. Unlike object storage, we have a specific service that only works for exoscale. Even the kind is explicitly `ExoscalePostgreSQL` (it's not possible or sensible to use a cloudscale or AWS specific DBaaS composition for a `ExoscalePostgreSQL` for example). I see no need to have multiple compositions for this XRD, and thus this field doesn't need to be configurable.

docs/modules/ROOT/pages/references/composite/dbaas/exoscale/postgres.adoc
```diff
@@ -0,0 +1,11 @@
...
+
+The parent key for all of the following parameters is`pkg.appcat.composite.dbaas.exoscale.postgres`.
+
+== `defaultCompositionRef`
+
+[horizontal]
+type:: string
+default:: `${facts:cloud}.exoscalepostgresqls.appcat.vshn.io`
+
+The default composition to use when a user creates a Claim of the composite resource.
```
ccremer, line 11: As mentioned earlier, there's no need for multiple compositions, this section can be removed
```suggestion
```

packages/composite/dbaas/exoscale/postgres.yml
```diff
@@ -0,0 +1,176 @@
...
+          connectionSecretKeys:
+            - POSTGRESQL_URL
+            - POSTGRESQL_DB
+            - POSTGRESQL_HOST
+            - POSTGRESQL_PORT
+            - POSTGRESQL_USER
+            - POSTGRESQL_PASSWORD
+            - ca.crt
+          defaultCompositionRef:
+            name: ${pkg.appcat.composite.exoscalepostgresql:defaultCompositionRef}
```
ccremer, line 29: From #32  this should be hardcoded
```suggestion
            name: exoscale.postgres.appcat.vshn.io
```

ccremer (changes requested):

packages/composite/dbaas/exoscale/postgres.yml
ZuGao: ah makes sense, we do not need for it another composition.

ccremer:
> But I'm fine with merging this as is and creating a tech debt to migrate later.

> Let's do a full migration in another PR. I started this PR before https://github.com/vshn/component-appcat/pull/33

I disagree here. We should generate the YAML as much as possible to keep it maintainable. IMO this is important enough to do it in this PR already and not later.

Kidswiss:
> I disagree here. We should generate the YAML as much as possible to keep it maintainable. IMO this is important enough to do it in this PR already and not later.

I've already created a tech debt that can be scheduled for the next sprint. IMHO that's the more iterative approach than to cram it into the same PR.

Kidswiss (approved):

PR #37, opened 2022-10-26
Add defaults to the postgres exoscale XRD

ZuGao:
* Add defaults to the composition

## Checklist

### DIFF
--- a/packages/composite/dbaas/exoscale/postgres.yml
+++ b/packages/composite/dbaas/exoscale/postgres.yml
@@ -45,6 +45,7 @@ parameters:
                                   description: MajorVersion is the major version identifier for the
                                     instance.
                                   type: string
+                                  default: "14"
                                 pgSettings:
                                   description: PGSettings contains additional PostgreSQL settings.
                                   type: object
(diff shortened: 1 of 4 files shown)


Kidswiss:
Okay I think didn't talk about the same thing 🗡️ 
My suggestion was to add it to the composite (XRD) not the composition, so it's visible via `kubectl explain`.

packages/composite/dbaas/exoscale/postgres.yml
```diff
@@ -45,6 +45,7 @@ parameters:
                                   description: MajorVersion is the major version identifier for the
                                     instance.
                                   type: string
+                                  default: 14
```
Kidswiss, line 48: You probably have to specify it as "14" becuase string :)

```suggestion
                                  default: "14"
```

Kidswiss (changes requested):
Just one small thing, the rest LGTM.

Have you tested if it can still be applied? Or if there's some error somewhere?

ZuGao:
Moved the defaults from composition to composite as suggested by @Kidswiss

Kidswiss (approved):

ZuGao:
> Have you tested if it can still be applied? Or if there's some error somewhere?

I only test it with golden tests. And yes the output looks good.

ccremer (approved):

Issue #15, opened 2022-08-24
Generate the XRDs using Go structs with KubeBuilder


ccremer:
## Context

Problem:
The XRD spec contains an API scheme (parameters, status etc) that is pretty much native Kubernetes OpenAPI scheme. Manually maintaining this scheme is error prone. I've looked around and didn't find a tool in https://openapi.tools/ that covers the case at hand.

Idea:
Kubernetes Operators today are built with Kubebuilder, a code generator for building CRDs. The CRDs also follow the OpenAPI scheme.
We could setup some boilerplate and setup the Types and Fields in Go structs (see https://github.com/vshn/provider-cloudscale/blob/f68758cc396358852ba09a0c9ff3b5a3ad877521/apis/cloudscale/v1/bucket_types.go) and let KubeBuilder/`go generate` generate the API.
That creates a CRD though, but with some yq/Go/jsonnet magic we can basically copy the generated scheme from the CRD into the XRD.

## ACs

* Setup Go code to generate a CRD using Kubebuilder (see example: https://github.com/vshn/provider-cloudscale/blob/master/apis/generate.go and related subdirectory)
* Once the CRD is created, merge/copy the OpenAPI scheme from the CRD (`.spec.versions[*]`) to the composite (`.spec.versions[*]`)
  * maybe with `yq` or a small Go script
* GitHub action that aborts a check when the generated composition is out-of-date.

Kidswiss:
Go-Swagger can create openapi from code: https://goswagger.io/use/spec/model.html

But it will also generate a lot of other stuff we don't need. So going through kubebuilder seems the better option IMHO.

ccremer:
I would only consider KubeBuilder, as that's the only generator that pretty much guarantees a scheme that Kubernetes understands. (for example, special flags for type handling, see https://book.kubebuilder.io/reference/markers/crd-processing.html)

As we've built a bunch of Operators by now, we have plenty of experience with Kubebuilder, so the initial setup should be fairly easy.

ccremer:
First implementation done, see https://github.com/vshn/component-appcat/blob/cef1e368fd00d0b848052b904066c43b4ad2ad69/apis/v1/objectstorage_types.go as an example

PR #38, opened 2022-11-01
Generate Postgresql DBaaS composite with Go

ccremer:
This PR introduces the generation of PostgreSQL on Exoscale XRD using Go structs. It should make it easier in the future to maintain the OpenAPI schema

Related: #15 

The diff is somewhat hard to read, but all the fields should stay, and there are a few changed default values.

## Checklist


ccremer (commented):
Note: I couldn't reuse the same Go types from provider-exoscale, because those types have conflicting kubebuilder comments.
For example, for provider-exoscale, certain fields are required and need to be set (e.g. backup and maintenance settings). However, for the XRD, those fields can be VSHN-optionated defaults.

To be able to reuse the same Go types, we'd have to remove most of the kubebuilder comments and move the validation logic into validation webhook on the provider-exoscale side, and we'd need to setup both validation and defaulter webhooks for the XRDs as well. This is out of scope here so I copied the same structs. Might be worth discussing in a later stage.

ZuGao (approved):
LGTM

PR #41, opened 2022-11-08
Add Redis DBaaS composite

mweibel:
Adds redis.

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/references/composite/dbaas/exoscale/redis.adoc
+++ b/docs/modules/ROOT/pages/references/composite/dbaas/exoscale/redis.adoc
@@ -1,0 +1,8 @@
+= pkg.appcat.composite.dbaas.exoscale.redis
+
+The parent key for all of the following parameters is `pkg.appcat.composite.dbaas.exoscale.redis`.
+
+[NOTE]
+====
+Currently, there's nothing to configure.
+====
--- a/packages/composite/dbaas/exoscale/redis.yml
+++ b/packages/composite/dbaas/exoscale/redis.yml
@@ -1,0 +1,147 @@
+classes:
+  - '...common'
+parameters:
+  appcat:
+    composites:
+      "exoscaleredis.appcat.vshn.io":
+        spec:
+          group: appcat.vshn.io
+          names:
+            kind: XExoscaleRedis
+            plural: xexoscaleredis
+          claimNames:
+            kind: ExoscaleRedis
+            plural: exoscaleredis
+          connectionSecretKeys:
+            - REDIS_HOST
+            - REDIS_PORT
+            - REDIS_USERNAME
+            - REDIS_PASSWORD
+            - REDIS_URL
+            - ca.crt
+          defaultCompositionRef:
+            name: exoscale.redis.appcat.vshn.io
+          versions:
+            - additionalPrinterColumns:
+                - jsonPath: .status.parameters.size.plan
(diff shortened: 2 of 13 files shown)


ccremer (approved):

docs/modules/ROOT/pages/references/composite/dbaas/exoscale/redis.adoc
```diff
@@ -0,0 +1,8 @@
+= pkg.appcat.composite.dbaas.exoscale.redis
+
+The parent key for all of the following parameters is`pkg.appcat.composite.dbaas.exoscale.redis`.
```
ZuGao, line 3:
```suggestion
The parent key for all of the following parameters is `pkg.appcat.composite.dbaas.exoscale.redis`.
```

packages/composite/dbaas/exoscale/redis.yml
```diff
@@ -0,0 +1,147 @@
...
+        spec:
+          group: appcat.vshn.io
+          names:
+            kind: XExoscaleRedis
+            plural: xexoscaleredis
+          claimNames:
+            kind: ExoscaleRedis
+            plural: exoscaleredis
+          connectionSecretKeys:
+            - host
```
ZuGao, line 16: Can we do all upper case or all lowercase? doesn't seem right to me to have this mixed.

ZuGao (commented):
Some small things.

packages/composite/dbaas/exoscale/redis.yml
mweibel: but that's how it's exposed in the connection secret.

I'm not sure either if it's that grave, TBH. An application might mount `REDIS_URL` directly as an env variable while another might individually get the other keys in some other way, depending on the use case? 
But no idea what is best here - happy to change but that would require a change in provider-exoscale, too.

docs/modules/ROOT/pages/references/composite/dbaas/exoscale/redis.adoc
mweibel: nice catch :+1: updated the otherr files too/

packages/composite/dbaas/exoscale/redis.yml
ZuGao: I would go ahead and update the provider and make it consistent with other configurations like [postgres](https://github.com/vshn/component-appcat/blob/master/packages/composite/dbaas/exoscale/postgres.yml#L16) or [mysql](https://github.com/vshn/component-appcat/blob/master/packages/composite/dbaas/exoscale/mysql.yml#L16).

ccremer: strictly speaking though, the ticket for provider-exoscale mentioned the variables in lower-case, that's why Michael left it according to the ticket.

ZuGao: I thought for a second that this might have been done on purpose during the provider-exoscale implementation. @ccremer do you think we should leave as is at this point?

ccremer: IMO we wait for https://github.com/vshn/provider-exoscale/pull/58 and incorporate the changes here.

mweibel: fixed and updated

ccremer (approved):

PR #46, opened 2022-11-11
Create common exoscale api-secret

ZuGao:
Currently we are providing multiple api-keys for provider exoscale for each service (dbaas and object-storage). This implies that the provider can work either with only one service or with a wrong vault reference. With this PR I'm moving the exoscale api secret management away from services and to the provider itself.

## Checklist


ccremer (approved):

packages/composition/dbaas/exoscale/postgres.yml
```diff
@@ -117,14 +117,6 @@ parameters:
...
 
-    secrets:
-      exoscale-dbaas-api-access:
-        metadata:
-          namespace: ${crossplane:namespace}
-        stringData:
-          EXOSCALE_API_KEY: '?{vaultkv:${cluster:tenant}/${cluster:name}/appcat/dbaas/provider-exoscale/access-key}'
-          EXOSCALE_API_SECRET: '?{vaultkv:${cluster:tenant}/${cluster:name}/appcat/dbaas/provider-exoscale/secret-key}'
-
   crossplane:
```
mweibel, line 120: is there a reason postgres has the provideConfig still?

PR #48, opened 2022-11-21
cloudscale metrics collector with k8s api support

### LINKED ISSUE #30: Get rid of the namespace observer cloudscale
## Context

Currently, for the ObjectBucket XRD and composition we use provider-kubernetes and its Object resource to watch the namespace of the claims. We need that to figure out the APPUiO Cloud organization name for billing purposes (see more: https://kb.vshn.ch/appuio-cloud/references/architecture/metering-data-flow.html). This organization information is in the namespace labels where the claim resides.

The current approach is to copy that `appuio.io/organization` label to the managed resource, so that the metrics collectors can pick up this information from the managed resources.

### Idea

it may be beneficial if the metrics collector directly query the Kubernetes API and figure out this namespace label themselves. We reduce complexity of the compositions, but only slightly increase complexity in the metrics collector. The collectors would have to simply GET the namespace objects for each managed resource. 

### Impact
* cloudscale-metrics-collector gets the organization from the Objects user metadata tags that the cloudscale API offers. This project would need the plumbing to query the organization from namespace instead of the tags. However, exoscale-metrics-collector has this plumbing already so should be fairly easy to adapt as well.
* The composition no longer provision provider-kubernetes Object resources

## Out of Scope

* Commodore component does not need to be updated yet.

## Alternatives

Continue using the current watch-and-patch approach.

## Acceptance Criteria

* Extend cloudscale-metrics-collector with: 
  * Boilerplate to setup Kubernetes client  
  * listing Namespaces and organization label (make this part reusable for exoscale-metrics-collector)
  * querying all managed resources
  * Stop using tags from cloudscale API
  * Compare objects from kubernetes with buckets from cloudscale API
* Adjust composition so that
  * it doesn't use provider-kubernetes anymore (and stops patching the ns label in managed resource)
  * it stops using tags for cloudscale API
* Test out the composition and collector on our Lab cluster

mweibel:
fixes #30 

## Checklist

### DIFF
--- a/packages/composition/objectstorage/cloudscale.yml
+++ b/packages/composition/objectstorage/cloudscale.yml
@@ -13,8 +13,6 @@ parameters:
         commonPatchSets:
           annotations: {}
           labels: {}
-        commonResources:
-          observeClaimNamespace: {}
         spec:
           compositeTypeRef:
             apiVersion: appcat.vshn.io/v1
@@ -68,14 +66,6 @@ parameters:
                     strategy: string
                     string:
                       fmt: "%s.%s"
-                # Set tenant tag
-                - type: FromCompositeFieldPath
-                  fromFieldPath: metadata.labels[appuio.io/organization]
-                  toFieldPath: spec.forProvider.tags.tenant
-                # Set namespace tag
-                - type: FromCompositeFieldPath
-                  fromFieldPath: metadata.labels[crossplane.io/claim-namespace]
-                  toFieldPath: spec.forProvider.tags.namespace
 
             - base:
                 apiVersion: cloudscale.crossplane.io/v1
--- a/packages/tests/golden/composition-objectstorage-cloudscale/appcat/appcat/compositions.yaml
+++ b/packages/tests/golden/composition-objectstorage-cloudscale/appcat/appcat/compositions.yaml
@@ -68,12 +68,6 @@ spec:
               - fromFieldPath: metadata.labels[crossplane.io/claim-name]
           toFieldPath: spec.forProvider.displayName
           type: CombineFromComposite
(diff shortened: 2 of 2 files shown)


packages/composition/objectstorage/cloudscale.yml
```diff
@@ -135,13 +127,6 @@ parameters:
                         rma: rma
                         lpg: lpg
 
-    additionalResources:
-      provider-cloudscale-secrets-objectsusers:
-        apiVersion: v1
-        kind: Namespace
-        metadata:
-          name: ${pkg.appcat.composition.objectstorage.cloudscale:secretNamespaces:objectsusers}
```
mweibel, line 143: I'm not sure if this should actually be removed or not?

Kidswiss: No this one is still necessary, it's the namespace where we save the intermediate secrets of the Managed Resources in the composition.

mweibel: ah yeah, that's what I thought. Removed it from the changes list again.

Kidswiss:
Also you can remove the whoie `kubernetes.crossplane.io` from https://github.com/vshn/component-appcat/blob/66c044a777dfe638c5711c98a12190e145daf580/packages/tests/golden/composition-objectstorage-cloudscale/appcat/appcat/compositions.yaml#L122 which is probably managed by: https://github.com/vshn/component-appcat/blob/2a589d3ddd248aec8bccc97bd77831181cf58ec4/packages/composition/objectstorage/cloudscale.yml#L17

mweibel:
ah, right. That's the one :) Removed.

ZuGao (approved):
LGTM

PR #50, opened 2022-12-05
Add default versions for mysql and postgres

ZuGao:
## Checklist


Kidswiss:
Are currently really only one version supported for MySQL and PostgreSQL?

tobru:
How do we keep that up-to-date?

ZuGao:
> Are currently really only one version supported for MySQL and PostgreSQL?

Yes only one major version.

glrf (changes requested):
If we want to do this, please set it through kubebuilder.

But tobru has a good point. I'm not certain anymore if exposing the version makes sense for exoscale, especially if all DBaaS only expose a single version anyway. 

And at least for kafka the only way to find out the available version was to try and see which one doesn't fail, so I don't see how we could automate that dependency update

Kidswiss:
> But tobru has a good point. I'm not certain anymore if exposing the version makes sense for exoscale, especially if all DBaaS only expose a single version anyway.

Well new major versions of these apps aren't released that often, so manually managing it for now seems reasonable.

glrf (approved):

PR #54, opened 2022-12-07
remove namespace observer exoscale

### LINKED ISSUE #35: Get rid of the namespace observer exoscale
## Context

Currently, for the ObjectBucket XRD and composition we use provider-kubernetes and its Object resource to watch the namespace of the claims. We need that to figure out the APPUiO Cloud organization name for billing purposes (see more: https://kb.vshn.ch/appuio-cloud/references/architecture/metering-data-flow.html). This organization information is in the namespace labels where the claim resides.

The current approach is to copy that `appuio.io/organization` label to the managed resource, so that the metrics collectors can pick up this information from the managed resources.

### Idea

it may be beneficial if the metrics collector directly query the Kubernetes API and figure out this namespace label themselves. We reduce complexity of the compositions, but only slightly increase complexity in the metrics collector. The collectors would have to simply GET the namespace objects for each managed resource. 

### Impact
* exoscale-metrics-collector already has the plumbing for querying Kubernetes API. Adding a query for the namespace should be trivial
* The composition no longer provision provider-kubernetes Object resources

## Out of Scope

* Commodore component does not need to be updated yet.

## Alternatives

Continue using the current watch-and-patch approach.

## Acceptance Criteria

* Extend exoscale-metrics-collector with listing Namespaces
* Adjust composition so that it doesn't use provider-kubernetes anymore (and stops patching the ns label in managed resource)
* Test out the composition and collector on our Lab cluster

susana-garcia:
Fixes #35 

## Checklist

- [X] Categorize the PR by setting a good title and adding one of the labels:

### DIFF
--- a/packages/composition/dbaas/exoscale/kafka.yml
+++ b/packages/composition/dbaas/exoscale/kafka.yml
@@ -12,8 +12,6 @@ parameters:
         commonPatchSets:
           annotations: {}
           labels: {}
-        commonResources:
-          observeClaimNamespace: {}
         spec:
           compositeTypeRef:
             apiVersion: appcat.vshn.io/v1
--- a/packages/composition/dbaas/exoscale/mysql.yml
+++ b/packages/composition/dbaas/exoscale/mysql.yml
@@ -12,8 +12,6 @@ parameters:
         commonPatchSets:
           annotations: {}
           labels: {}
-        commonResources:
-          observeClaimNamespace: {}
         spec:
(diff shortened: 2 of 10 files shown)


glrf (approved):
LGTM.

Can be merged as soon as https://github.com/vshn/exoscale-metrics-collector/pull/31 is merged, released and rolled out.

ZuGao (approved):
LGTM

PR #55, opened 2022-12-09
Fix nested defaulting for XRDs

glrf:
Kubebuilder cannot set empty objects as a default for CRDs. We might be able to fix this upstream, but I'll do this temporary hack to fix this here.

This could be done with less duplication, but as we want to move away from packages anyway, we should soon be able to do this in a cleaner way in Jsonnet.
So, I don't think investing more time to make this prettier is helpful at this point.

Conflicts with #57 and #56. Whoever merges later will need to rebase and fix this.

## Checklist

### DIFF
--- a/apis/v1/dbaas_exoscale.go
+++ b/apis/v1/dbaas_exoscale.go
@@ -25,7 +25,7 @@ type ExoscaleDBaaSMaintenanceScheduleSpec struct {
 }
 
 type ExoscaleDBaaSSizeSpec struct {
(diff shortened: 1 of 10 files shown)


susana-garcia:
LGTM, I guess once this is on master, we can inform the ones with open PRs to rebase their own branches.

glrf:
I think yours shouldn't clash. But @TheBigLee  will need to make sure to update the generate commands and @wejdross will need to add the same generate commands for opensearch

PR #56, opened 2022-12-12
Add Composite and Composition for OpenSearch

wejdross:
## Checklist


packages/tests/sample.yaml
```diff
@@ -0,0 +1,21 @@
...
+#    backup:
+#      timeOfDay: "12:00:00"
+#    maintenance:
+#      timeOfDay: "12:10:00"
+#      dayOfWeek: "tuesday"
+#    service:
+#      version: "2"
+#      zone: "ch-gva-2"
+#    size:
+#      plan: "startup-4"
```
glrf, line 21: Could you remove that. We don't have any samples for the other DBaaS and it definitely shouldn't be in the test directory.

glrf (commented):
LGTM overall, some smaller things

* Please add the added test cases to github actions here: https://github.com/vshn/component-appcat/blob/master/.github/workflows/packages.yaml#L17-L54
* Please add reference docs for the new packages here: https://github.com/vshn/component-appcat/tree/master/docs/modules/ROOT/pages/references
* Please update the title to something helpful like "Add Composite and Composition for OpenSearch", as the title will show up in the release notes

susana-garcia:
LGTM, I haven't test it locally

susana-garcia (approved):
apart from the changes that Fabian pointed out, everything looks good

packages/tests/sample.yaml
wejdross: sure

glrf: Thanks, but please also update the make target as it tries to apply a manifest that doesn't exist anymore

PR #68, opened 2023-01-03
Refactor to a single self contained Component

glrf:
This is a huge refactor, but there are only very few changes to the resulting resources, they're just shuffled around. I migrated on the Lab cluster by selectively syncing in ArgoCD, and the only actual changes are:

* Providers and their config is moved from the crossplane App to the appcat App
* The exoscale secrets namespace is now deployed once instead of 6 times
* The object storage and dbaas composition now use the same provider config, instead of two separate ones that are exactly the same

I will add the component to https://github.com/projectsyn/commodore-component-template in a separate PR after merging.

If you want to review the changes in detail, i tried to make each commit reviewable in itself  by updating the component and the remaining package in sync. And you can always ping me and we can look through it together.

## Checklist


susana-garcia (approved):
just a small improve, the rest LGMT, but I didn't test it locally.
I review intensively the jsonnet files in `\component` and couldn't find any error (bravo!).
Also thank you for the laborious work.

README.md
```diff
@@ -27,11 +23,11 @@ $ cd component
 $ make test
```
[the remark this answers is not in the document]
glrf: Oh totally missed that, thanks!

apis/generate.mk
```diff
@@ -1,9 +1,10 @@
 clean_targets += .clean-apis
 
 .PHONY: generate-xrd
-generate-xrd: ## Generates the XRDs using Kubebuilder
+generate-crd: ## Generates the XRDs using Kubebuilder
```
zugao, line 4: Aren't we generating composites here? If that's the case I would leave XRD instead of CRD. CRD is a broader term.

class/appcat.yml
```diff
@@ -0,0 +1,21 @@
+parameters:
+  kapitan:
+    compile:
```
zugao, line 3: Is there a reason we don't add all dependencies right away? such as `crossplane.libsonnet` that is needed for providers?

docs/modules/ROOT/pages/references/provider-exoscale.adoc
```diff
@@ -1,11 +1,52 @@
...
+
+== `connectionSecretNamespace`
+[horizontal]
+type:: string
+default:: `syn-provider-exoscale-secrets`
+
+A namespace to store the connection secrets in.
+The namespace will be created and managed by this component.
 
+== `credentials`
```
zugao, line 20: I would add here the variable names and maybe how you can them.

docs/modules/ROOT/pages/references/provider-cloudscale.adoc
```diff
@@ -1,11 +1,52 @@
...
+A namespace to store the connection secrets in.
+The namespace will be created and managed by this component.
 
+== `credentials`
 [horizontal]
 type:: dict
-default:: see `packages/provider/cloudscale.yml`
 
-The package image repository for https://github.com/vshn/provider-cloudscale[provider-cloudscale].
+Credentials to connect to the cloudscale.ch API.
```
zugao, line 24: I would add here the variable names and maybe how you can them.

docs/modules/ROOT/pages/references/service-objectstorage.adoc
```diff
@@ -0,0 +1,80 @@
+= Objecstorage Service
```
zugao, line 1: I would add here the reason of why `generic` dict exists.

zugao (approved):
The refactoring is a huge step forward from what he had initially and since I already worked several days with this code base I can proudly say that we succeeded big time on this repo. The code is clean, we get well defined separation of concerns, the composition patches layout is great and removes a lot of bloat code from crossplane. I was able to easily understand the code and extend the functions from various libs. I added some small concerns that you might want to take a look but overall LGTM.

apis/generate.mk
glrf: Yeah naming is hard.  Technically the command now generated CRDs that are then turned into XRDs in Jsonnet. So I renamed the command. But I guess this confuses more than it helps..

zugao: Then I would leave `generate-crd` and add more details in the comment.

class/appcat.yml
glrf: Yes, because that's how commodore handles library dependencies. We shouldn't add them explicitly as they'll be present if the corresponding component is installed on the cluster. We only need to add the dependencies in the tests as there this doesn't work.

Jsonnet library dependencies in commodore aren't as good as they could be yet, so for now we're stuck with it.

PR #74, opened 2023-01-17
Add network policy

zugao:
* Added network policy
* Fixed a bug with pacthing
* Updated SGInstanceProfile to suit our needs

## Checklist


glrf (commented):
LGTM overall, but I'd suggest to make it possible to disable the network policy creation.

If a cluster does not come with any default network polices, but is open by default, adding this policy will block all traffic not originating from the claim namespace (For example also from prometheus).

glrf (approved):

PR #75, opened 2023-01-18
Add billing and organization labels for VSHN managed PostgreSQL

glrf:
This PR adds billing as well as organization labels to the created namespace for PostgeSQL.

## Checklist

### DIFF
--- a/component/vshn_postgres.jsonnet
+++ b/component/vshn_postgres.jsonnet
@@ -219,12 +219,25 @@ local composition =
       ],
       resources: [
         {
+          base: namespace {
+            spec+: {
+              managementPolicy: 'Observe',
+            },
+          },
+          patches: [
+            comp.FromCompositeFieldPathWithTransformPrefix('metadata.labels[crossplane.io/composite]', 'metadata.name', 'ns-observer'),
+            comp.FromCompositeFieldPath('metadata.labels[crossplane.io/claim-namespace]', 'spec.forProvider.manifest.metadata.name'),
+            comp.ToCompositeFieldPath('status.atProvider.manifest.metadata.labels[appuio.io/organization]', 'metadata.labels[appuio.io/organization]'),
+          ],
+        },
+        {
           base: namespace,
           patches: [
             comp.PatchSetRef('annotations'),
             comp.PatchSetRef('labels'),
             comp.FromCompositeFieldPathWithTransformPrefix('metadata.labels[crossplane.io/composite]', 'metadata.name', 'vshn-postgresql'),
             comp.FromCompositeFieldPath('metadata.labels[crossplane.io/claim-namespace]', 'spec.forProvider.manifest.metadata.labels[%s]' % serviceNamespaceLabelKey),
+            comp.FromCompositeFieldPath('metadata.labels[appuio.io/organization]', 'spec.forProvider.manifest.metadata.labels[appuio.io/organization]'),
           ],
         },
         {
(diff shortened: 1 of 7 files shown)


component/vshn_postgres.jsonnet
```diff
@@ -211,12 +218,26 @@ local composition =
...
       resources: [
+        {
+          base: namespace {
+            spec+: {
+              managementPolicy: 'Observe',
+            },
+          },
+          patches: [
+            comp.FromCompositeFieldPathWithTransformPrefix('metadata.labels[crossplane.io/composite]', 'metadata.name', 'ns-observer'),
+            comp.FromCompositeFieldPath('metadata.labels[crossplane.io/claim-namespace]', 'spec.forProvider.manifest.metadata.name'),
```
zugao, line 229: I don't understand this patch. The label `crossplane.io/claim-namespace` is literally the name of this observed namespace. Also the policy is Observe therefore you cannot patch anything in this namespace resource, right?

component/vshn_postgres.jsonnet
```diff
@@ -211,12 +218,26 @@ local composition =
...
+            comp.ToCompositeFieldPath('status.atProvider.manifest.metadata.labels[appuio.io/organization]', 'metadata.labels[appuio.io/organization]'),
+          ],
+        },
         {
           base: namespace,
           patches: [
             comp.PatchSetRef('annotations'),
             comp.PatchSetRef('labels'),
             comp.FromCompositeFieldPathWithTransformPrefix('metadata.labels[crossplane.io/composite]', 'metadata.name', 'vshn-postgresql'),
+            comp.FromCompositeFieldPath('metadata.labels[crossplane.io/claim-namespace]', 'spec.forProvider.manifest.metadata.labels[%s]' % serviceNamespaceLabelKey),
```
zugao, line 239: Interesting to know you can do this kind of trick :)

component/vshn_postgres.jsonnet
```diff
@@ -17,6 +17,9 @@ local defaultUser = 'postgres';
 local defaultPort = '5432';
 
 
+local serviceNameLabelKey = 'appcat.vshn.io/servicename';
+local serviceNamespaceLabelKey = 'appcat.vshn.io/claim-namespace';
```
zugao, line 21: I'm a bit confused with the name of this label. Why do we use `claim-namespace` here? Would it make sense to add `service` word in the label?

zugao (commented):
Everything seems good but I am a bir confused by some naming and patches.

component/vshn_postgres.jsonnet
glrf: Maybe I missunderstood how the observer should work. But this basically tells crossplane which namespace it needs to observer right?

component/vshn_postgres.jsonnet
glrf: I don't have strong feelings about that name. We just need to somehow have a label that tells us in which namespace this corresponding claim is for billing. `claim-namespace` seemed like an obvious choice, but we can use whatever we want

component/vshn_postgres.jsonnet
zugao: Ah sry, I miss calculated, you are still patching the provider kubernetes objects. Everything good.

component/vshn_postgres.jsonnet
zugao: I thought thoroughly about this label and realised the name is actually what we want. Maybe I could add a word that the purpose of this label is for the billing. Now, not sure if you want to include that in the group or in the name of the label.

zugao (approved):
LGTM

PR #76, opened 2023-01-18
Allow provider kubernetes to delete created resources

glrf:
Currently the provider kubernetes can create multiple resources, but can't delete them. This causes the deployed VSHN postgres instance to stay around, even if the claim is deleted.

## Checklist


zugao (approved):
LGTM, and I was wondering why the namespace would not be deleted :)

PR #81, opened 2023-01-24
Fix connection secrets for VshnPostgreSQL

TheBigLee:
Setting the parameter `writeConnectionSecretToRef` on the claim should
result in a secret with the connection credentials named after the value
set in the `name` key of that parameter. However, the current
implementation writes the connection credentials to a hardcoded secret
called `<claim-name>-connection`. This leads to confusions for the end
user. This commit will fix this issue, by creating a temporary secret in
the namespace where the instance is placed and will then copy the
content to the secret specified in the claim.

Signed-off-by: Nicolas Bigler <nicolas.bigler@vshn.ch>

## Checklist

      changelog.
      linked in the changelog.


glrf (approved):
LGTM,  you just need to run `make gen-golden-all` then this looks good to me

glrf (approved):

zugao (approved):
LGTM. It's cool to see this already fixed considering that just yesterday evening we were only coming out with possible solutions.

PR #84, opened 2023-01-31
Appcat 268

wejdross:
## Checklist

      changelog.
      linked in the changelog.


wejdross:
I have this weird issue with pipeline, I tried rebasing, but only messed it up more, so please review

class/defaults.yml
```diff
@@ -57,6 +57,10 @@ parameters:
         enabled: false
         secretNamespace: ${crossplane:namespace}
         postgres:
+          # bucket_region: 'lpg' || 'ch-gva-2'
+          bucket_region: ''
+          # bucket_endpoint: 'https://objects.lpg.cloudscale.ch' || 'https://sos-ch-gva-2.exo.io'
+          bucket_endpoint: ''
```
glrf, line 63: Please fail compilation if these aren't set.

glrf (changes requested):
Also please add documentation for the added paramters

class/defaults.yml
wejdross: I added assertions but now more things fails, it works when You isolate test case:
`make test -e instance=vshn` 

but fails when `make gen-golden-all` is run

glrf: That's because you assert this unconditionally. You need to only assert this if PostgreSQL by VSHN is actually enabled

apis/vshn/v1/dbaas_vshn_postgresql.go
```diff
@@ -108,14 +109,15 @@ type VSHNDBaaSNetworkSpec struct {
 	IPFilter []string `json:"ipFilter,omitempty"`
 }
 
-// VSHNDBaaSBackupSpec contains settings to control the backups of an instance.
-type VSHNDBaaSBackupSpec struct {
-	// +kubebuilder:validation:Pattern="^([0-1]?[0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])$"
-	// +kubebuilder:default="21:30:00"
+type VSHNPostgreSQLBackup struct {
+	// +kubebuilder:validation:Pattern=^(\*|([0-9]|1[0-9]|2[0-9]|3[0-9]|4[0-9]|5[0-9])|\*\/([0-9]|1[0-9]|2[0-9]|3[0-9]|4[0-9]|5[0-9])) (\*|([0-9]|1[0-9]|2[0-3])|\*\/([0-9]|1[0-9]|2[0-3])) (\*|([1-9]|1[0-9]|2[0-9]|3[0-1])|\*\/([1-9]|1[0-9]|2[0-9]|3[0-1])) (\*|([1-9]|1[0-2])|\*\/([1-9]|1[0-2])) (\*|([0-6])|\*\/([0-6]))$
+	// +kubebuilder:default=* * * * *
```
zugao, line 114: Do we want to run it every minute as default?

apis/vshn/v1/dbaas_vshn_postgresql.go
```diff
@@ -108,14 +109,15 @@ type VSHNDBaaSNetworkSpec struct {
...
+	Schedule string `json:"schedule,omitempty"`
 
-	// TimeOfDay for doing daily backups, in UTC.
-	// Format: "hh:mm:ss".
-	TimeOfDay string `json:"timeOfDay,omitempty"`
+	// +kubebuilder:validation:Pattern="^[1-9][0-9]*$"
+	// +kubebuilder:default=6
+	// +kubebuilder:validation:XIntOrString
```
zugao, line 119: Since this is clearly an `Int` I would remove `String` validation.

class/defaults.yml
```diff
@@ -57,6 +57,10 @@ parameters:
         enabled: false
         secretNamespace: ${crossplane:namespace}
         postgres:
+          # bucket_region: 'lpg' || 'ch-gva-2'
+          bucket_region: ''
```
zugao, line 61: Do we really need the region here? If I'm not wrong there's only one endpoint for buckets for Exoscale and Cloudscale. The region is necessary. You might want to check that though.

class/defaults.yml
zugao: I would also consider to create an object `backup` where all backup info would reside. Also consider if we can enable or disable backups altogether.

component/vshn_postgres.jsonnet
```diff
@@ -119,6 +118,14 @@ local composition =
                               sgInstanceProfile: '',
                               configurations: {
                                 sgPostgresConfig: '',
+                                backups:
+                                  [
+                                    {
+                                      sgObjectStorage: '',
+                                      cronSchedule: '0 22 * * *',
```
zugao, line 125: Why do you expose the Schedule in the claim meanwhile you hardcode the value in the composition? Am I missing something?

docs/modules/ROOT/pages/references/services-vshn.adoc
```diff
@@ -42,3 +42,28 @@ type:: bool
...
+
+.Examples
+* Cloudscale
+** 'lpg'
+* Exoscale
+** 'ch-gva-2' 
+
+=== `postgres.bucket_endpoint`
+type:: string
+default:: `''`
```
zugao, line 61: Maybe we want a default here @Kidswiss ? I know that in some cases we used Cloudscale as default.

zugao (commented):
Some things to check, overall good.

apis/vshn/v1/dbaas_vshn_postgresql.go
wejdross: will be fixed, should be 0 22

apis/vshn/v1/dbaas_vshn_postgresql.go
wejdross: this is not a mistake if I try to leave pure int I receive:
```
component-appcat/apis/vshn/v1/dbaas_vshn_postgresql.go:119:2: must apply pattern to a `string` or `IntOrString`
```

class/defaults.yml
wejdross: yes we need, I spoke about it with Simon and Fabian and they all want it to be cloud specific, so I need to make it parameter that will be specified in project syn for deployment

component/vshn_postgres.jsonnet
wejdross: will be fixed, thanks for noticing

wejdross:
@zugao @glrf merge ready?

zugao:
@wejdross there are still opened discussions.

class/defaults.yml
wejdross: I agree, but if we actually have at least one use case of postgresql without backup

Kidswiss: We probably can't disable backups easily without any conditionals in the composition. Let's stick with fix enabled backups for now.

docs/modules/ROOT/pages/references/services-vshn.adoc
Kidswiss: The default should be whatever the underlying cloud is.

E.g. APPUiO Cloudscale -> cloudscale
APPUiO AWS -> aws
etc.

Not sure if we can do dynamic facts as defaults?

zugao: I see we have cloud in the [facts](https://syn.tools/commodore/reference/parameters.html#_facts). @wejdross you may try to use this as default.

class/defaults.yml
zugao: Fine by me.

docs/modules/ROOT/pages/references/services-vshn.adoc
wejdross: @glrf can You put an input here, because as I spoke with You about that, You pointed out that might not be the best idea to go with facts::cloud

glrf: We can't find the bucket endpoint from the facts, we'd need to hardcode bucket endpoints per cloud. Then deploying it to a new cloud would involve adding new hardcoded endpoints.

IMO the way we currently do it is more generic and not really more effort. Once we have appcat on more than three clusters such config can be added to the global defaults.

zugao: @wejdross please next time just add a comment earlier as to why the idea we discussed (me and @Kidswiss ) is not the best way to move forward. I feel this discussion was dragged too long here.

component/vshn_postgres.jsonnet
```diff
@@ -290,6 +360,38 @@ local composition =
             comp.FromCompositeFieldPathWithTransformSuffix('metadata.labels[crossplane.io/claim-name]', 'spec.writeConnectionSecretToRef.name', 'connection'),
           ],
         },
+        // s3 bucket creation for backup purposes
+        {
+          base: objectBucket,
```
Kidswiss, line 365: The `s3-bucket` suffixes and pre-fixes seem somewhat redundant here. You create an `ObjectBucket` object, so it's clear what it's doing. The `s3-bucket` suffix doesn't provide more information here. 

Also, we'll end up with buckets that start with `s3-bucket-*`, which is also not very helpful when browsing the buckets.

Kidswiss (changes requested):

component/vshn_postgres.jsonnet
Kidswiss: And something else:

Please use the `XObjectBucket` (composite) and not the `ObjectBucket` (claim). The claims are not intended to be used in compositions.

The `XObjectBucket` behaves like any other Crossplane managed resource and is cluster scoped. So, you can then also remove the namespace as well.

PR #86, opened 2023-02-02
Fix image asset location and document postgres resource usage and pgsettings more explicit

TheBigLee:
## Checklist

      changelog.
      linked in the changelog.


apis/vshn/v1/dbaas_vshn_postgresql.go
```diff
@@ -84,12 +86,12 @@ type VSHNDBaaSMaintenanceScheduleSpec struct {
 type VSHNDBaaSSizeSpec struct {
 	// +kubebuilder:default="500m"
 
-	// CPU defines the amount of Kubernetes CPUs for an instance.
+	// CPU defines the amount of Kubernetes CPUs for an instance (will set requests and limits to this value).
```
tobru, line 89: Is it not possible to set request and limits separately? This way it collides with the resource ratio constraint of APPUiO Cloud quite hard, and it can cause high compute costs for a user because requests are directly billed.

TheBigLee: I have opened an issue for this: #87

TheBigLee: Just checked the docs for stackgres. They explicitly discourage separating requests and limits for production workload and requires setting a specifc configuration option to enable this: https://stackgres.io/doc/latest/reference/crd/sgcluster/#non-production-options
Before we implement this, we should properly discuss if we want to do this.

PR #88, opened 2023-02-06
Add VSHNManaged Redis support

TheBigLee:
## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/apis/vshn/v1/dbaas_vshn_redis.go
+++ b/apis/vshn/v1/dbaas_vshn_redis.go
@@ -1,0 +1,79 @@
+package v1
+
+import (
+	v1 "github.com/vshn/component-appcat/apis/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Workaround to make nested defaulting work.
+// kubebuilder is unable to set a {} default
+//go:generate yq -i e ../../generated/vshn.appcat.vshn.io_vshnredis.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.default={})"
+//go:generate yq -i e ../../generated/vshn.appcat.vshn.io_vshnredis.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.properties.size.default={})"
+//go:generate yq -i e ../../generated/vshn.appcat.vshn.io_vshnredis.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.properties.service.default={})"
+
+// +kubebuilder:object:root=true
+
+// VSHNRedis is the API for creating Redis clusters.
+type VSHNRedis struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec defines the desired state of a VSHNRedis.
+	Spec VSHNRedisSpec `json:"spec"`
+}
+
+// VSHNRedisSpec defines the desired state of a VSHNRedis.
+type VSHNRedisSpec struct {
+	// Parameters are the configurable fields of a VSHNRedis.
+	Parameters VSHNRedisParameters `json:"parameters,omitempty"`
+}
+
+// VSHNRedisParameters are the configurable fields of a VSHNRedis.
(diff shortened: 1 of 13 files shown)


apis/vshn/v1/dbaas_vshn_redis.go
```diff
@@ -0,0 +1,101 @@
...
+	MemoryLimits string `json:"memoryLimits,omitempty"`
+
+	// +kubebuilder:default="5Gi"
+
+	// Disk defines the amount of disk space for an instance.
+	Disk string `json:"disk,omitempty"`
+}
+
+// VSHNRedisBackupSpec contains settings to control the backups of an instance.
+type VSHNRedisBackupSpec struct {
```
Kidswiss, line 88: I'd probably omit the backup config for now, as far as I can see the chart doesn't bring any backups with it. So we'll still have to engineer it and figure out how to handle the configuration.

apis/vshn/v1/dbaas_vshn_redis.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// TimeOfDay for doing daily backups, in UTC.
+	// Format: "hh:mm:ss".
+	TimeOfDay string `json:"timeOfDay,omitempty"`
+}
+
+// VSHNRedisStatus reflects the observed state of a VSHNRedis.
+type VSHNRedisStatus struct {
+	// RedisConditions contains the status conditions of the backing object.
+	RedisConditions []v1.Condition `json:"redisConditions,omitempty"`
```
Kidswiss, line 100: I don't see a patch in the composition that would actually fill in these conditions.

Kidswiss (changes requested):
Just two remarks, the rest LGTM. But haven't tested it locally.

apis/vshn/v1/dbaas_vshn_redis.go
TheBigLee: True, since we haven't implemented anything yet. However, we will probably need it soon to get some events forwarded to the claim. Should I remove it for now or leave it?

apis/vshn/v1/dbaas_vshn_redis.go
TheBigLee: Removed it

apis/vshn/v1/dbaas_vshn_redis.go
Kidswiss: Leaving it in without populating it could lead to confusions... I'd remove it until it's properly added.

apis/vshn/v1/dbaas_vshn_redis.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// VSHNRedisServiceSpec contains Redis DBaaS specific properties
+type VSHNRedisServiceSpec struct {
+	// +kubebuilder:validation:Enum="6.2";"7.0"
+	// +kubebuilder:default="7.0"
+
+	// Version contains supported version of Redis.
+	// Multiple versions are supported. The latest version "7.0" is the default version.
+	Version string `json:"version,omitempty"`
```
glrf, line 50: If I get this right this version value is used directly as the image tag. So we'll always update patch versions on restart.

I think this is fine for the first iteration, but we'll need to think about how we handle this in the future, I'd rather have proper version pins here and not update it whenever.

glrf (approved):
LGTM overall

apis/vshn/v1/dbaas_vshn_redis.go
TheBigLee: Yes correct. We will have to think how to handle this later on

apis/vshn/v1/dbaas_vshn_redis.go
TheBigLee: Ok. removed it as well

class/defaults.yml
```diff
@@ -64,6 +77,12 @@ parameters:
           enabled: true
           enableNetworkPolicy: true
           secretNamespace: ${appcat:services:vshn:secretNamespace}
+        redis:
+          enabled: true
+          enableNetworkPolicy: true
+          secretNamespace: ${appcat:services:vshn:secretNamespace}
+          helmChartVersion: 17.6.0
```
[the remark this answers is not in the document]
TheBigLee: Good point. Fixed

PR #93, opened 2023-02-09
Don't pass the annotations and labels to the managed resources

### LINKED ISSUE #92: Don't allow to pass the `external-name` annotation from the claim to the managed resource
Currently, all annotations from the claim are forwarded and applied to the managed resource. This includes the `crossplane.io/external-name` annotation. With this annotation you can adopt any instance running in the same account.
This is a big security issue as you can hijack any instance if you know or guess the instance name and gain full control of the instance as you will get the credentials for that instance and can access all data. Furthermore, you can change the instance settings and even delete the instance completely.

We therefore need to disable the forwarding of the `crossplane.io/external-name` annotation from the claim to the managed resource.

## Steps to Reproduce the Problem

  1. Create for example a redis instance by following: https://docs.appuio.cloud/appcat/exoscale-dbaas/redis/create.html
  1. Get the external name of the instance with `k get redis.exoscale.crossplane.io -A`
  1. Create another redis instance but set the annotation `crossplane.io/external-name` to the external name of the other instance
  1. Connection to the other instance is now possible with the secret and also changing settings is possible. Also deleting the claim will delete the initial instance.

## Actual Behavior
- Hijacking of any instance currently possible

## Expected Behavior
- Setting the `crossplane.io/external-name` annotation doesn't do anything and is completely ignored

Kidswiss:
* Removed the patchsets that pass all annotations and labels
* Removed patchset functions as they provide not much benefit when using jsonnet

This removes the user's ability to set any annotations and labels on the managed resources. This resolves security problems with the `external-name`.

Resolves #92

## Checklist

      changelog.
      linked in the changelog.
- [X] Update the documentation.

### DIFF
--- a/component/exoscale_postgres.jsonnet
+++ b/component/exoscale_postgres.jsonnet
@@ -71,17 +71,11 @@ local composition =
     spec: {
       compositeTypeRef: comp.CompositeRef(xrd),
       writeConnectionSecretsToNamespace: pgParams.secretNamespace,
-      patchSets: [
-        comp.PatchSet('annotations'),
-        comp.PatchSet('labels'),
-      ],
       resources: [
         {
           base: pgBase,
           connectionDetails: comp.conn.AllFromSecretKeys(connectionSecretKeys),
           patches: [
-            comp.PatchSetRef('annotations'),
-            comp.PatchSetRef('labels'),
             comp.FromCompositeFieldPath('metadata.labels[crossplane.io/composite]', 'metadata.name'),
             comp.FromCompositeFieldPath('metadata.labels[crossplane.io/composite]', 'spec.writeConnectionSecretToRef.name'),
 
--- a/component/exoscale_kafka.jsonnet
+++ b/component/exoscale_kafka.jsonnet
@@ -67,17 +67,11 @@ local composition =
     spec: {
       compositeTypeRef: comp.CompositeRef(xrd),
       writeConnectionSecretsToNamespace: kafkaParams.secretNamespace,
-      patchSets: [
-        comp.PatchSet('annotations'),
-        comp.PatchSet('labels'),
-      ],
       resources: [
         {
           base: kafkaBase,
           connectionDetails: comp.conn.AllFromSecretKeys(connectionSecretKeys),
           patches: [
-            comp.PatchSetRef('annotations'),
-            comp.PatchSetRef('labels'),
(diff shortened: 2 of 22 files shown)


wejdross:
looks ok

TheBigLee (approved):
LGTM

component/exoscale_postgres.jsonnet
```diff
@@ -71,17 +71,11 @@ local composition =
...
-        comp.PatchSet('labels'),
-      ],
       resources: [
         {
           base: pgBase,
           connectionDetails: comp.conn.AllFromSecretKeys(connectionSecretKeys),
           patches: [
-            comp.PatchSetRef('annotations'),
-            comp.PatchSetRef('labels'),
             comp.FromCompositeFieldPath('metadata.labels[crossplane.io/composite]', 'metadata.name'),
```
glrf, line 79: AFAIK the billing collector depends on the label `crossplane.io/claim-namespace` being set. See: https://github.com/vshn/billing-collector-cloudservices/blob/master/pkg/exoscale/dbaas.go#L175

We'll need to explicitly forward that. Otherwise I think this should be fine

glrf (changes requested):

component/exoscale_postgres.jsonnet
Kidswiss: The `crossplane.io/*` labels are autogenerated. The only thing that the removed patchsets forward are user set annotations and labels on the claim. I'm not really sure why we did this in the first place...

After creating an `evil` intance without the patchsets there are still the labels on the namespaces:

```
apiVersion: v1
kind: Namespace
metadata:
  labels:
    appcat.vshn.io/claim-namespace: default
    appcat.vshn.io/servicename: postgresql-standalone
    kubernetes.io/metadata.name: vshn-postgresql-evil-m89lq
  name: vshn-postgresql-evil-m89lq
```

Kidswiss: Looks like those labels are autopopulated to managed resources as well:

```
k get postgresqls.exoscale.crossplane.io my-postgres-example-27898 -o jsonpath='{.metadata.labels}' | jq
{
  "crossplane.io/claim-name": "my-postgres-example",
  "crossplane.io/claim-namespace": "default",
  "crossplane.io/composite": "my-postgres-example-27898"
}
```

There's no need to explicitly patch them.

glrf (approved):

PR #96, opened 2023-02-15
Add feature to specify a nodeSelector for the postgreSQL instance

TheBigLee:
## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/apis/vshn/v1/dbaas_vshn_postgresql.go
+++ b/apis/vshn/v1/dbaas_vshn_postgresql.go
@@ -44,6 +44,9 @@ type VSHNPostgreSQLParameters struct {
 	// Size contains settings to control the sizing of a service.
 	Size VSHNDBaaSSizeSpec `json:"size,omitempty"`
 
+	// Scheduling contains settings to control the scheduling of an instance.
+	Scheduling VSHNDBaaSSchedulingSpec `json:"scheduling,omitempty"`
+
 	// Network contains any network related settings.
 	Network VSHNDBaaSNetworkSpec `json:"network,omitempty"`
 
@@ -62,6 +65,12 @@ type VSHNPostgreSQLServiceSpec struct {
 
 	// PGSettings contains additional PostgreSQL settings.
 	PostgreSQLSettings runtime.RawExtension `json:"pgSettings,omitempty"`
+}
+
+// VSHNDBaaSSchedulingSpec contains settings to control the scheduling of an instance.
+type VSHNDBaaSSchedulingSpec struct {
+	// NodeSelector is a selector which must match a node’s labels for the pod to be scheduled on that node
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 }
 
 // VSHNDBaaSMaintenanceScheduleSpec contains settings to control the maintenance of an instance.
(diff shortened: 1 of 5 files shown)


apis/vshn/v1/dbaas_vshn_postgresql.go
```diff
@@ -64,6 +67,12 @@ type VSHNPostgreSQLServiceSpec struct {
 	PostgreSQLSettings runtime.RawExtension `json:"pgSettings,omitempty"`
 }
 
+// VSHNPostgreSQLSchedulingSpec contains settings to control the scheduling of an instance.
+type VSHNPostgreSQLSchedulingSpec struct {
```
zugao, line 71: Isn't the name a bit ambiguous? The customer may think it's PostgreSQL related.

glrf (approved):

glrf:
This looks good to me. My only question is if we really want to expose the whole node selector. Only exposing the node class would probably be a better UX on APPUiO Cloud (but won't make sense once we are on APPUiO Managed so :shrug: )

TheBigLee:
> This looks good to me. My only question is if we really want to expose the whole node selector. Only exposing the node class would probably be a better UX on APPUiO Cloud (but won't make sense once we are on APPUiO Managed so shrug )

I think it makes sense to expose the whole node selector, since the end-user has to specify the whole node selector on his pod or namespace as well. In this way the behavior is almost consistent.
The most consistent way however, would be to have a `spec.nodeSelector` in the claim and patching that to the `sgCluster` object.
What do you think?

apis/vshn/v1/dbaas_vshn_postgresql.go
TheBigLee: Renamed it to `VSHNDBaaSSchedulingSpec`. Does that make more sense?

apis/vshn/v1/dbaas_vshn_postgresql.go
```diff
@@ -64,6 +67,12 @@ type VSHNPostgreSQLServiceSpec struct {
 	PostgreSQLSettings runtime.RawExtension `json:"pgSettings,omitempty"`
 }
 
+// VSHNDBaaSSchedulingSpec contains settings to control the scheduling of an instance.
+type VSHNDBaaSSchedulingSpec struct {
+	// NodeSelector is a selector which must match a node’s labels for the pod to be scheduled on that node
+	NodeSelector runtime.RawExtension `json:"nodeSelector,omitempty"`
```
glrf, line 73:
```suggestion
	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```

I didn't see that last time. I'd suggest to be more restrictive and use the same type as upstream: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3191-L3197

glrf (approved):

glrf:
> I think it makes sense to expose the whole node selector, since the end-user has to specify the whole node selector on his pod or namespace as well. In this way the behavior is almost consistent.
> The most consistent way however, would be to have a `spec.nodeSelector` in the claim and patching that to the `sgCluster` object.
> What do you think?

I think it's fine as is right now. You're right it's more consistent for the user to specify the whole `nodeSelector`.

apis/vshn/v1/dbaas_vshn_postgresql.go
zugao: yes, don't forget to update the second PR as well.

zugao (approved):
LGTM

PR #97, opened 2023-02-16
Local dev S3 composition

Kidswiss:
This commit adds a composition that implements our XObjectBucket XRD. It will provision a local minio instance and a bucket. Currently only supports creating one instance and bucket.

This relates to the implementation of the PostgreSQL backups, as the backup depends on the ObjectBucket.

## Checklist

      changelog.
      linked in the changelog.
- [X] Update the documentation.

### DIFF
--- a/component/provider.jsonnet
+++ b/component/provider.jsonnet
@@ -8,7 +8,6 @@ local crossplane = import 'lib/crossplane.libsonnet';
 local inv = kap.inventory();
 local params = inv.parameters.appcat;
 
-
 local addCredentials(config, credentials) = config {
   credentials: std.get(config, 'credentials', default=credentials),
 };
--- a/dev/component/objectstorage.jsonnet
+++ b/dev/component/objectstorage.jsonnet
@@ -1,0 +1,182 @@
+local com = import 'lib/commodore.libjsonnet';
+local kap = import 'lib/kapitan.libjsonnet';
+local kube = import 'lib/kube.libjsonnet';
+
+local comp = import 'lib/appcat-compositions.libsonnet';
+local crossplane = import 'lib/crossplane.libsonnet';
+
+local common = import '../component/common.libsonnet';
+local xrds = import '../component/xrds.libsonnet';
+
+local inv = kap.inventory();
+local params = inv.parameters.appcat;
+local objStoParams = params.services.generic.objectstorage;
+
+local xrd = xrds.XRDFromCRD(
+  'xobjectbuckets.appcat.vshn.io',
+  xrds.LoadCRD('appcat.vshn.io_objectbuckets.yaml'),
+  defaultComposition='%s.objectbuckets.appcat.vshn.io' % objStoParams.defaultComposition,
+  connectionSecretKeys=[
+    'AWS_ACCESS_KEY_ID',
+    'AWS_SECRET_ACCESS_KEY',
+    'AWS_REGION',
+    'ENDPOINT',
+    'ENDPOINT_URL',
+    'BUCKET_NAME',
+  ]
+);
+
+local minioRbac =
+  local provider = params.providers.helm;
+
+  local sa = kube.ServiceAccount(provider.controllerConfig.serviceAccountName) {
+    metadata+: {
+      namespace: provider.namespace,
+    },
+  };
+  local role = kube.ClusterRole('crossplane:provider:provider-helm:system:dev') {
+    rules: [
+      {
+        apiGroups: [ '' ],
+        resources: [ 'persistentvolumeclaims', 'deployments' ],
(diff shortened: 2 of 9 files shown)


Kidswiss:
I'm not sure if this is the right approach to have a composition that is only intended for local development.

class/defaults.yml
```diff
@@ -113,6 +113,12 @@ parameters:
           secretNamespace: ${appcat:services:exoscale:secretNamespace}
           providerSecretNamespace: ${appcat:services:exoscale:providerSecretNamespace}
 
+      # Dev services for local development
+      dev:
+        # deploys a simple minio composition to provide local objectbuckets
+        # this composition implements the objectbucket xrd
+        enableLocalObjectBucket: false
+
```
glrf, line 121: I'd rather not have any defaults related to local development. It invites people to try them on a cluster..

component/objectstorage.jsonnet
```diff
@@ -241,9 +242,124 @@ local compositionExoscale =
...
+          ],
+          patches: [
+            comp.FromCompositeFieldPath('metadata.labels[crossplane.io/composite]', 'spec.forProvider.values.buckets[0].name'),
+            comp.FromCompositeFieldPath('metadata.labels[crossplane.io/composite]', 'spec.writeConnectionSecretToRef.name'),
+          ],
+        },
+      ],
+    },
+  };
+
```
glrf, line 358: I'd rather not mix "test-code" and actual code. I'd suggest to move this to `tests/component/objectstorage.jsonnet` or something like that.

component/provider.jsonnet
```diff
@@ -217,6 +217,22 @@ local controllerConfigRef(config) =
...
+        {
+          apiGroups: [ 'apps' ],
+          resources: [ 'deployments' ],
+          verbs: [ 'get', 'list', 'watch', 'create', 'watch', 'patch', 'update', 'delete' ],
+        },
+        {
+          apiGroups: [ 'batch' ],
+          resources: [ 'jobs' ],
+          verbs: [ 'get', 'list', 'watch', 'create', 'watch', 'patch', 'update', 'delete' ],
+        },
```
glrf, line 235: Same here. I don't like this being in the actual code base

tests/vshn.yml
```diff
@@ -5,6 +5,9 @@ parameters:
         source: https://raw.githubusercontent.com/projectsyn/component-crossplane/v2.3.0/lib/crossplane.libsonnet
         output_path: vendor/lib/crossplane.libsonnet
 
+  facts:
+    cloud: dev
+
```
glrf, line 10: Instead of matching on a cloud `dev` I'd add a separate kapitan compile step. Then there is no chance that this will end up in production

glrf (changes requested):
I see where this is coming from, but as this is right now I'm very concerned that this will end up in production somehow. Also I'd very much prefer to have a clean separation of test code and actual code.

class/defaults.yml
zugao: second that

zugao (commented):
I'm not sure if I like to see a dev composition in this repo. Can we find an alternative solution?

Kidswiss:
> I'm not sure if I like to see a dev composition in this repo. Can we find an alternative solution?

I've implemented Fabian's suggestions. The dev component is now separate from the productive ones.

component/objectstorage.jsonnet
```diff
@@ -11,6 +11,7 @@ local xrds = import 'xrds.libsonnet';
 local inv = kap.inventory();
 local params = inv.parameters.appcat;
 local objStoParams = params.services.generic.objectstorage;
+local devParams = params.services.dev;
```
glrf, line 14: We don't need that anymore

component/provider.jsonnet
```diff
@@ -7,7 +7,7 @@ local crossplane = import 'lib/crossplane.libsonnet';
 
 local inv = kap.inventory();
 local params = inv.parameters.appcat;
-
+local devParams = params.services.dev;
```
glrf, line 10: Same here

dev/component/objectstorage.jsonnet
```diff
@@ -0,0 +1,186 @@
...
+            comp.FromCompositeFieldPath('metadata.labels[crossplane.io/composite]', 'spec.writeConnectionSecretToRef.name'),
+          ],
+        },
+      ],
+    },
+  };
+
+if objStoParams.enabled then {
+  '20_xrd_objectstorage': xrd,
+  '20_rbac_objectstorage': xrds.CompositeClusterRoles(xrd),
```
glrf, line 183: You're overwriting the XRD and rbac of the actual component. I wouldn't do that.

glrf (approved):
LGTM overall

PR #98, opened 2023-02-16
Add PostgreSQL Restore Ability

Kidswiss:
This PR adds the ability to do a restore from an existing instance.
To do so it does following actions:

* deploy a clusterRole, namespace (appcat-control) and serviceaccount
* create a new composition for restores, which is 99% the same as the
  normal one
* the composition creates in turn a job to copy over the necessary information from the source
  instance. This happens in the appcat-control namespace.
* a new sgCluster is created that points to the copied objects for the
  restore parameters

The rest is handled by StackGres itself.

## Checklist

      changelog.
      linked in the changelog.
- [] Update the documentation.

### DIFF
--- a/class/defaults.yml
+++ b/class/defaults.yml
@@ -96,6 +96,8 @@ parameters:
           enabled: true
           enableNetworkPolicy: true
           secretNamespace: ${appcat:services:vshn:secretNamespace}
+          # Used for deploying jobs during restores
+          controlNamespace: 'syn-appcat-control'
         redis:
           enabled: true
           enableNetworkPolicy: true
--- a/component/scripts/copy-pg-backup.sh
+++ b/component/scripts/copy-pg-backup.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+set -e
+
+xrdname=$(kubectl -n "${CLAIM_NAMESPACE}" get vshnpostgresqls "${CLAIM_NAME}" -ojson | jq -r '.spec.resourceRef.name')
+
+source_namespace=$(kubectl -n "${CLAIM_NAMESPACE}" get vshnpostgresqls "${CLAIM_NAME}" -ojson | jq -r '.status.instanceNamespace')
+
+echo "copy secret"
+kubectl -n "${source_namespace}" get secret "pgbucket-${xrdname}" -ojson | jq 'del(.metadata.namespace) | del(.metadata.ownerReferences)' | kubectl -n "${TARGET_NAMESPACE}" apply -f -
+echo "copy sgObjectStorage"
+kubectl -n "${source_namespace}" get sgobjectstorages.stackgres.io "sgbackup-${xrdname}" -ojson | jq 'del(.metadata.namespace) | del(.metadata.ownerReferences)' | kubectl -n "$TARGET_NAMESPACE" apply -f -
+echo "copy sgBackup"
+kubectl -n "${source_namespace}" get sgbackups.stackgres.io "${BACKUP_NAME}" -ojson | jq '.spec.sgCluster = .metadata.namespace + "." + .spec.sgCluster | del(.metadata.namespace) | del(.metadata.ownerReferences)' | kubectl -n "${TARGET_NAMESPACE}" apply -f -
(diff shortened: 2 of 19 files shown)


Kidswiss:
As discussed, the clusterRole is not optimal. But I haven't found a viable solution that works without it and makes crossplane happy.

Also you can test it locally:
```
# install stackgres helm and certmanager on kind
# then install the golden files
k apply -f tests/golden/vshn/appcat/appcat/

first instance:

cat <<EOF | kubectl apply -f -
apiVersion: vshn.appcat.vshn.io/v1
kind: VSHNPostgreSQL
metadata:
  name: pgsql-app1-prod
spec:
  parameters:
    backup:
      schedule: '*/5 * * * *'
    service:
      majorVersion: "15"
      pgSettings:
        timezone: Europe/Zurich
    size:
      cpu: "600m"
      memory: "3500Mi"
      disk: "80Gi"
  writeConnectionSecretToRef:
    name: postgres-creds
EOF

second instance:
cat <<EOF | kubectl apply -f -
apiVersion: vshn.appcat.vshn.io/v1
kind: VSHNPostgreSQL
metadata:
  name: pgsql-app2-prod
spec:
  parameters:
    restore:
      claimName: pgsql-app1-prod
      backupName: # k get sgbackups --all-namespaces
    backup:
      schedule: '0 22 * * *'
    service:
      majorVersion: "15"
      pgSettings:
        timezone: Europe/Zurich
    size:
      cpu: "600m"
      memory: "3500Mi"
      disk: "80Gi"
  writeConnectionSecretToRef:
    name: postgres-creds-restored
  compositionRef:
    name: vshnpostgresrestore.vshn.appcat.vshn.io
EOF

```

Unfortunately the local ObjectBucket provider currently only supports a single bucket. In order for the second instance to provision correctly, you need to comment out the xobjectbucket and sgobjectstorage from the restore composition. As well as remove the backup key and patches from the sgcluster object.

I was able to successfully restore a database with data in it. I haven't yet tested the point in time recovery though.

zugao (approved):
for me looks good, I haven't tested it.

Kidswiss:
I've implemented the suggested change to put the job into a separate namespace.

While it's a bit ugly it's probably the best solution we can currently achieve given the limitations.

component/scripts/copy-pg-backup.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -e
+
+xrdname=$(kubectl -n "${CLAIM_NAMESPACE}" get vshnpostgresqls "${CLAIM_NAME}" -ojson | jq -r '.spec.resourceRef.name')
+
+source_namespace=$(kubectl -n "${CLAIM_NAMESPACE}" get vshnpostgresqls "${CLAIM_NAME}" -ojson | jq -r '.status.instanceNamespace')
+
+echo "copy secret"
+kubectl -n "${source_namespace}" get secret "pgbucket-${xrdname}" -ojson | jq 'del(.metadata.namespace)' | kubectl -n "${TARGET_NAMESPACE}" apply -f -
```
zugao, line 10: Where the `TARGET_NAMESPACE` comes from?

glrf (approved):
LGTM overall, just some nitpicks

component/vshn_postgres.jsonnet
```diff
@@ -41,508 +41,649 @@ local xrd = xrds.XRDFromCRD(
...
-                                [serviceNamespaceLabelKey]: '',
-                              },
-                            },
-                          },
-                        },
-                      },
-                    };
-  local localca = comp.KubeObject('cert-manager.io/v1', 'Issuer') +
+local controlnamespaceName = 'appcat-control';
+local controlNamespace = kube.Namespace(controlnamespaceName);
```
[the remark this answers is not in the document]
Kidswiss: I'll keep it in mind 👍

component/scripts/copy-pg-backup.sh
Kidswiss: It gets patched in the composition.

class/defaults.yml
```diff
@@ -80,6 +80,8 @@ parameters:
           enabled: true
           enableNetworkPolicy: true
           secretNamespace: ${appcat:services:vshn:secretNamespace}
+          # Used for deploying jobs during restores
+          controlNamespace: 'appcat-control'
```
glrf, line 84: Please add this to the parameter docs

glrf (approved):
LGTM

PR #101, opened 2023-02-17
Fix defaults to remove ArgoCD diff

glrf:
Crossplane is a bit special with when it adds defaults to fields and when it doesn't.  This change should fix the ArgoCD sees because of set defaults.
## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/vshn_postgres.jsonnet
+++ b/component/vshn_postgres.jsonnet
@@ -333,9 +333,7 @@ local composition =
     {
       apiVersion: 'appcat.vshn.io/v1',
       kind: 'XObjectBucket',
-      metadata: {
-        name: '',
-      },
+      metadata: {},
       spec: {
         parameters: {
           bucketName: '',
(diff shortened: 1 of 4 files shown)


wejdross:
LGTM

component/vshn_postgres.jsonnet
```diff
@@ -333,9 +333,7 @@ local composition =
     {
       apiVersion: 'appcat.vshn.io/v1',
       kind: 'XObjectBucket',
-      metadata: {
-        name: '',
-      },
+      metadata: {},
```
Kidswiss, line 336: Nit: from a crossplane perspective if `metadata.name` gets patched anyway you can omit the whole object.

No idea what would happen with the argo diff then.

wejdross (approved):
LGTM

Kidswiss (approved):
Just a small nit, rest LGTM

component/vshn_postgres.jsonnet
glrf: Hmm not sure about the argodiff. Currently it reports a `metadata: {}` so i took that. Both would probably work.

Kidswiss: Feel free to leave as is, it's just a nit :)

PR #106, opened 2023-03-01
Allow setting resource requests and limits separately for VshnPostgreSQL

### LINKED ISSUE #87: Allow setting resource requests and limits separately for VshnPostgreSQL
## Context
Currently, it's not possible to set the resource requests and limits separately for a VSHNPostgreSQL instance.
This can cause high compute costs for a user because requests are directly billed and are set to the same as the limit, forcing users to set high values for the resource requests even if they don't need it all the time.

Separating the requests and limits allows the user to safe money and also doesn't end up blocking the cluster with lots of idle instances.

## ACs
* `nonProductionOptions` enabled
* `Refactor 

## Implementation ideas
* Enable the following `nonProductionOptions` on the `SGCluster` object:
  * `enableSetPatroniCpuRequests`
  * `enableSetPatroniMemoryRequests`
* Refactor the existing fields `memory` and `cpu` to `memoryRequests`, `memoryLimits`, `cpuRequests`, and `cpuLimits` and map them to the corresponding `SGInstanceProfile` fields:
  * `spec.cpu` (for limits)
  * `spec.requests.cpu` (for requests)
  * `spec.memory` (for limits)
  * `spec.requests.memory` (for limits)
 * By default set the `requests` and `limits` to the same value by adhering the platforms ration (e.g. `500m` CPU and `2000Mi` memory)
 * Document the default values in the end-user documentation

## Alternatives
* Provide fixed plans for the end-user

TheBigLee:
Currently, it's not possible to set resource requests to different values then the request limits.

This PR adds new parameters to be able to set the resource requests explicitly.

This resolves #87 

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/apis/vshn/v1/dbaas_vshn_postgresql.go
+++ b/apis/vshn/v1/dbaas_vshn_postgresql.go
@@ -105,10 +105,22 @@ type VSHNDBaaSSizeSpec struct {
 	// Memory defines the amount of memory in units of bytes for an instance.
 	Memory string `json:"memory,omitempty"`
 
+	// Requests defines CPU and memory requests for an instance
+	Requests VSHNDBaaSSizeRequestsSpec `json:"requests,omitempty"`
+
 	// +kubebuilder:default="5Gi"
 
 	// Disk defines the amount of disk space for an instance.
 	Disk string `json:"disk,omitempty"`
+}
+
+// VSHNDBaaSSizeRequestsSpec contains settings to control the resoure requests of a service.
+type VSHNDBaaSSizeRequestsSpec struct {
+	// CPU defines the amount of Kubernetes CPUs for an instance.
+	CPU string `json:"cpu,omitempty"`
+
+	// Memory defines the amount of memory in units of bytes for an instance.
+	Memory string `json:"memory,omitempty"`
 }
 
 // VSHNDBaaSNetworkSpec contains any network related settings.
--- a/component/vshn_postgres.jsonnet
+++ b/component/vshn_postgres.jsonnet
@@ -204,6 +204,10 @@ local sgInstanceProfile = {
                 spec: {
                   cpu: '',
                   memory: '',
+                  requests: {
+                    cpu: null,
+                    memory: null,
+                  },
                   containers: {
                     'backup.create-backup': {
                       cpu: '250m',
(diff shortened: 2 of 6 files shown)


wejdross (approved):
Didn't test, but looks good

apis/vshn/v1/dbaas_vshn_postgresql.go
```diff
@@ -105,12 +105,28 @@ type VSHNDBaaSSizeSpec struct {
...
+type VSHNDBaaSSizeRequestsSpec struct {
+	// +kubebuilder:default="600m"
+
+	// CPU defines the amount of Kubernetes CPUs for an instance.
+	CPU string `json:"cpu,omitempty"`
+
+	// +kubebuilder:default="3500Mi"
+
+	// Memory defines the amount of memory in units of bytes for an instance.
+	Memory string `json:"memory,omitempty"`
```
glrf, line 127: Do we actually want to set defaults for these. If I understand the `SGInstanceProfile` correctly setting to request basically means that they're the same as the limit. Which is IMO more intuitive than using a default.

component/vshn_postgres.jsonnet
```diff
@@ -204,6 +204,10 @@ local sgInstanceProfile = {
                 spec: {
                   cpu: '',
                   memory: '',
+                  requests: {
+                    cpu: null,
+                    memory: null,
+                  },
```
glrf, line 210: Did you double check that this doesn't create a diff in ArgoCD, if no requests are set?

glrf (approved):
I wouldn't set defaults. Otherwise LGTM

component/vshn_postgres.jsonnet
TheBigLee: Yes, there is no diff. However, the manifest on the cluster is `requests: {}`.
I would still like to keep it thought for readability purposes as we do patch the `cpu` and `memory` fields in the composition.

apis/vshn/v1/dbaas_vshn_postgresql.go
TheBigLee: It's actually the other way around. Setting the limits and leaving the requests empty will set the requests to the same value as the limits. But I agree with you. I removed the defaults.

PR #111, opened 2023-03-06
Add patch maintenance script for PostgreSQL

Kidswiss:
This PR adds security maintenance to the PostgreSQL by VSHN service. The maintenance will be run on a schedule.

## Checklist

      changelog.
      linked in the changelog.


Kidswiss:
As mentioned, the actual conversion from dayOfWeek and timeOfDay is a bit ugly, but works, if the environmnet-config alpha feature is enabled.

component/vshn_postgres.jsonnet
```diff
@@ -638,6 +638,212 @@ local clusterRestoreConfig = {
...
+];
+
+local maintenanceJob = {
+  base: comp.KubeObject('batch/v1', 'CronJob') + {
+    spec+: {
+      forProvider+: {
+        manifest+: {
+          metadata+: {
+            namespace: pgParams.controlNamespace,
+          },
```
glrf, line 804: Nitpick: This is a bit misleading as we patch it anyways and it actually shouldn't run in the control namespace

glrf (approved):
LGTM. A bit hacky, but I don't have a cleaner solution

component/vshn_postgres.jsonnet
Kidswiss: Whoops, you're right, that's probably a copy paste error. Good catch.

PR #113, opened 2023-03-16
Add instance namespace permissions

Kidswiss:
Adds permissions for the enduser to the instance namespaces.

With this the user can:

- view events
- do portforwardings to pods
- read pod logs

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/main.jsonnet
+++ b/component/main.jsonnet
@@ -50,7 +50,23 @@ local finalizerRole = kube.ClusterRole('crossplane:appcat:finalizer') {
 
 };
 
+local readServices = kube.ClusterRole('appcat:services:read') + {
+  rules+: [
+    {
+      apiGroups: [ '' ],
+      resources: [ 'pods', 'pods/log', 'pods/status', 'events', 'services' ],
+      verbs: [ 'get', 'list', 'watch' ],
+    },
+    {
+      apiGroups: [ '' ],
+      resources: [ 'pods/portforward' ],
+      verbs: [ 'get', 'list', 'create' ],
+    },
+  ],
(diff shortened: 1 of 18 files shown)


component/main.jsonnet
```diff
@@ -50,7 +50,23 @@ local finalizerRole = kube.ClusterRole('crossplane:appcat:finalizer') {
...
+local readServices = kube.ClusterRole('appcat:services:read') + {
+  rules+: [
+    {
+      apiGroups: [ '' ],
+      resources: [ 'pods', 'pods/log', 'pods/status', 'events' ],
+      verbs: [ 'get', 'list', 'watch' ],
+    },
+    {
+      apiGroups: [ '' ],
+      resources: [ 'pods/portforward' ],
```
glrf, line 62: I might be a bit too cautious, but I'm not entirely sure that we want that. That way a customer can port-forward to an arbitrary (sidecar) container and complain because certain things don't work because they're holding it wrong (or worse maybe somehow break something)

Wouldn't port-forwarding to service be enough and would do what a customer actually what

glrf (approved):

component/main.jsonnet
glrf: Weird. I thought this should work: https://stackgres.io/doc/1.0/administration/cluster/connection/exposed/  but it somehow really doesn't :thinking:

Kidswiss: yeah, that's what I figured as well, I guess the docs aren't quite up to date.

But I'm adding port-forward for services as well.

PR #118, opened 2023-03-20
Add Prometheus rule for storage

Kidswiss:
This will alert on PostgreSQL storage filling up.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/vshn_postgres.jsonnet
+++ b/component/vshn_postgres.jsonnet
@@ -841,6 +841,59 @@ local maintenanceJob = {
   ] + convertToCron(),
 };
 
+local prometheusRule = {
+  base: comp.KubeObject('monitoring.coreos.com/v1', 'PrometheusRule') + {
+    spec+: {
+      forProvider+: {
+        manifest+: {
+          metadata: {
+            name: 'postgresql-storage-rules',
+          },
+          spec: {
+            groups: [
+              {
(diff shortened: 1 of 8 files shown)


component/vshn_postgres.jsonnet
```diff
@@ -841,6 +841,59 @@ local maintenanceJob = {
...
+          },
+          spec: {
+            groups: [
+              {
+                name: 'postgresql-storage',
+                rules: [
+                  {
+                    alert: 'PostgreSQLPersistentVolumeFillingUp',
+                    annotations: {
+                      description: 'The PersistentVolume claimed by {{ $labels.persistentvolumeclaim\n              }} in Namespace {{ $labels.namespace }} is only {{ $value |\n              humanizePercentage }} free.',
```
glrf, line 860: Does this actually work with a newline in the template?

glrf (approved):
Nitpick: Instead of notifying the customer that PVC X is running full in namespace Y, it would be a better UX if we could tell him that his database with claimname `foo` is running full. But I'm not sure if we can actually achieve this easily using PromQL

component/vshn_postgres.jsonnet
Kidswiss: No idea if opsgenie will respect them, but in the applied yaml file on the cluster they are there.

PR #131, opened 2023-04-03
Add AppCat SLI exporter

glrf:
Deploys the AppCat SLI exporter. Uses the commodore kustomize support https://syn.tools/syn/explanations/commodore-components/kustomizations.html

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/class/appcat.yml
+++ b/class/appcat.yml
@@ -2,7 +2,7 @@ parameters:
   kapitan:
     dependencies:
       - type: git
-        output_path: dependencies/appcat/manifests/${appcat:images:apiserver:tag}
+        output_path: dependencies/appcat/manifests/apiserver/${appcat:images:apiserver:tag}
         source: https://github.com/vshn/appcat-apiserver.git
         subdir: config
         ref: ${appcat:images:apiserver:tag}
@@ -28,3 +28,17 @@ parameters:
           - ${_base_directory}/component/vshn_redis.jsonnet
         input_type: jsonnet
         output_path: appcat/
+
+      - input_paths:
+          - ${_base_directory}/component/appcat_sli_exporter.jsonnet
+        input_type: jsonnet
+        output_path: ${_base_directory}/.work/appcat_sli_exporter
+        output_type: yaml
+      - input_type: external
+        output_path: .
+        input_paths:
+          - ${_kustomize_wrapper}
+        env_vars:
+          INPUT_DIR: ${_base_directory}/.work/appcat_sli_exporter
+        args:
+          - \${compiled_target_dir}/appcat/sli_exporter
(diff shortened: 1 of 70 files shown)


class/appcat.yml
```diff
@@ -28,3 +28,17 @@ parameters:
           - ${_base_directory}/component/vshn_redis.jsonnet
         input_type: jsonnet
         output_path: appcat/
+
+      - input_paths:
+          - ${_base_directory}/component/appcat_sli_exporter.jsonnet
+        input_type: jsonnet
+        output_path: ${_base_directory}/.work/appcat_sli_exporter
+        output_type: yaml
+      - input_type: external
```
zugao, line 37: Now I see where kustomize went

class/defaults.yml
```diff
@@ -48,6 +52,20 @@ parameters:
           cpu: 200m
           memory: 200Mi
 
+    slos:
```
zugao, line 55: This is an acronym for? sli exporter?

class/defaults.yml
```diff
@@ -48,6 +52,20 @@ parameters:
...
+      enabled: true
+      namespace: appcat-slos
```
zugao, line 57: we should probably need to add `syn-` so that alerts work. I had to do the same for the appcat api server.

docs/modules/ROOT/pages/references/slo-parameters.adoc
```diff
@@ -0,0 +1,31 @@
...
+[horizontal]
+type:: dict
+
+The resource requests and limits for the SLI Exporter.
+
+== `sli_exporter.kustomize_input`
+[horizontal]
+type:: dict
+
+The input passed to the Kustomize renderer.
```
zugao, line 30: I don't understand why this is necessary, can you explain the necessity of this parameter here? Why not use the namespace from the parent object directly?

zugao (commented):
Looks good but have some questions and remarks

class/defaults.yml
glrf: Service Level Objectives? The intention would be that we'll add other SLO relevant parameters later on

class/defaults.yml
glrf: Alerts will work regardless. But yes we'll need to change the namespace on APPUiO Cloud so that we don't alert Aldebaran. But I thought I'd be consistent with the other namespaces in the component defaults

docs/modules/ROOT/pages/references/slo-parameters.adoc
glrf: This just exposes more option to configure kustomize, similarly to how we usually expose the complete helm values. This is what is recommended by https://syn.tools/syn/explanations/commodore-components/kustomizations.html

zugao (approved):

PR #132, opened 2023-04-04
Fixing race condition in first reconciliation 

wejdross:
## Checklist

      changelog.
      linked in the changelog.


wejdross:
Context:
When we create empty instance then during first reconciliation loop pods are created without proper nodeSelector which is assigned by appuio=cloud-agent, that leads to race condition between patroni and openshift and as a result we have unstable solution - we need to restart pod to apply correct organisations and nodeSelector

By setting default organisation to `vshn` via simple label, we avoid that situation, in second reconciliation loop correct organisation label is taken from parent namespace

glrf (approved):
LGTM (but really isn't an enhancement :sweat_smile: ).

This _might_ be something we'll need to look at again once we support APPUiO Managed as this will add a useless/confusing label, but for now this should fix the issue and is good enough

PR #134, opened 2023-04-11
Add deletion protection to the VSHN PostgreSQL Instance

Kidswiss:
## Checklist

      changelog.
      linked in the changelog.


apis/vshn/v1/dbaas_vshn_postgresql.go
```diff
@@ -147,6 +155,16 @@ type VSHNPostgreSQLBackup struct {
 	// +kubebuilder:default=6
 	// +kubebuilder:validation:XIntOrString
 	Retention int `json:"retention,omitempty"`
+
+	// DeletionProtection will protect the instance from being deleted for the given retention time.
+	// This is enabled by default.
+	// +kubebuilder:default=true
```
glrf, line 161: Should we really default to `true`?

Honest question I don't know. I see that we want sane default settings for prod, but I think it's a bit unexpected to have this by default.

glrf (approved):
I'm not sure if we should enable deletion protection by default. Otherwise LGTM

apis/vshn/v1/dbaas_vshn_postgresql.go
zugao: I believe this should be set by default :)

PR #137, opened 2023-04-13
 Add UpdateStrategy to VSHNPostgreSQL

glrf:
We add an `updateStrategy` that works similarly to the statefulset
updateStrategy. The updateStrategy comes with two types

 - OnDelete: How we and stackgres currently handels this. Any changes to
   the spec that impact the workload such as changes to the resource
   requests, will only apply when the instance is restarted through
   other means, for example because of maitenance
 - OnChange: This will restart the instance whenever there is a change
   that requires a restart to apply. This is the new behaviour that
   allows service users to change their instance wihout having to wait
   for the next maintenance.

See https://github.com/vshn/appcat-comp-functions/pull/23 for related composition changes

## Checklist

      changelog.
      linked in the changelog.


apis/vshn/v1/dbaas_vshn_postgresql.go
```diff
@@ -74,6 +74,24 @@ type VSHNPostgreSQLParameters struct {
...
+	UpdateStrategy VSHNPostgreSQLUpdateStrategy `json:"updateStrategy,omitempty"`
+}
+
+const VSHNPostgreSQLUpdateStrategyTypeOnChange = "OnChange"
+const VSHNPostgreSQLUpdateStrategyTypeOnDelete = "OnDelete"
+
+// VSHNPostgreSQLUpdateStrategy indicates how and when updates the the instance spec will be applied.
+type VSHNPostgreSQLUpdateStrategy struct {
+	// +kubebuilder:validation:Enum="OnDelete";"OnChange"
+	// +kubebuilder:default="OnDelete"
```
glrf, line 88: I'm not entirely sure which behavior should be the default. What do you think?

zugao: I would definitely go with the safest route which is OnDelete. We do not want to surprise the customer.

zugao (commented):
I would actually argue a bit with the naming. What if we use `Now` and `Later`. Also if we keep the current proposal I would go with `OnRestart` instead of `OnDelete`. With `delete` keyword it seems a bit confusing, why would my changes apply when I delete my instance (from customer perspective).

Update: 
With keyword `Now` it's super straightforward, with keyword `Later` the customer will be forced to check the description which is already very well explained.

glrf:
> I would actually argue a bit with the naming. What if we use `Now` and `Later`. Also if we keep the current proposal I would go with `OnRestart` instead of `OnDelete`. With `delete` keyword it seems a bit confusing, why would my changes apply when I delete my instance (from customer perspective).

Yeah naming is hard..

I'm not really a fan of `Now` and `Later`. I initially had `Immediate` but changed it to `OnChange` and `OnDelete` as the StatefulSet has a field with the same name and has the type `RollingUpdate` and `OnDelete`. But I see your point.

For now I think `OnRestart` and `OnChange` would be my preferred option, but I'm not 100% happy with that either

apis/vshn/v1/dbaas_vshn_postgresql.go
glrf: Yeah, but that changing the disk size doesn't apply until maintenance is also a surprise :sweat_smile:

zugao:
> I'm not really a fan of `Now` and `Later`. I initially had `Immediate` but changed it to `OnChange` and `OnDelete` as the StatefulSet has a field with the same name and has the type `RollingUpdate` and `OnDelete`. But I see your point.

I expected this argument with K8s  convention. It's hard to argue against it but it's also sad that we cannot go with a more straightforward approach because of it. From my side the rest is ok so approved.

zugao (approved):

apis/vshn/v1/dbaas_vshn_postgresql.go
Kidswiss: I'd actually argue to go with the `OnChange` method by default. As it's how vanilla k8s does things as well and should be less of a surprise for users.

Kidswiss:
> For now I think `OnRestart` and `OnChange` would be my preferred option, but I'm not 100% happy with that either

My idea: `OnMaintenance` and `OnChange` this seems to be the least confusing naming for these arguments. At least from my side.

PR #139, opened 2023-04-21
Add Namespace field in objectreference

wejdross:
Why this change:
- when I deployed composition functions to openshift I discovered thanks to @glrf that we're missing Namespace field in struct which blocked creation of resurces -> reference to non existing field in composition && can't set it in claim because k8s doesn't know that field

### DIFF
--- a/apis/v1/objectreference.go
+++ b/apis/v1/objectreference.go
@@ -8,5 +8,6 @@ package v1
 type LocalObjectReference struct {
 	// Name of the referent.
 	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
-	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
+	Name      string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,1,opt,name=namespace"`
 }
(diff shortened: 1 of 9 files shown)


apis/v1/objectreference.go
```diff
@@ -8,5 +8,6 @@ package v1
 type LocalObjectReference struct {
 	// Name of the referent.
 	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
-	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
+	Name      string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,1,opt,name=namespace"`
```
glrf, line 12: We don't technically need this, but when we keep this we should do it correctly:
```suggestion
	Namespace string `json:"namespace,omitempty" protobuf:"bytes,2,opt,name=namespace"`
```

apis/v1/objectreference.go
```diff
@@ -8,5 +8,6 @@ package v1
 type LocalObjectReference struct {
```
glrf, line 8: Technically not a *Local*ObjectReference anymore, but changing this will be a breaking change so let's leave that. But please update the comment above, it's wrong now.

glrf (approved):
LGTM overall, but please try to add a description in the PR why we're doing this.

apis/v1/objectreference.go
wejdross: commited it separately

apis/v1/objectreference.go
wejdross: updated description field

PR #149, opened 2023-05-15
Add OpenShift templates to AppCat

Kidswiss:
## Summary

These OpenShift templates add integration into OpenShifts catalog. With this it's possible to provision simple AppCat instances right from the OpenShift console.

While testing this I also found, that Exoscale bumped their Kafka version and `3.2` isn't supported anymore.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/crds/exoscale.appcat.vshn.io_exoscalekafkas.yaml
+++ b/crds/exoscale.appcat.vshn.io_exoscalekafkas.yaml
@@ -90,7 +90,6 @@ spec:
                           default: "3.4"
                           description: Version contains the minor version for Kafka. Currently only "3.4" is supported. Leave it empty to always get the latest supported version.
                           enum:
-                            - "3.3"
                             - "3.4"
                           type: string
                         zone:
--- a/crds/exoscale.appcat.vshn.io_exoscalekafkas.yaml
+++ b/crds/exoscale.appcat.vshn.io_exoscalekafkas.yaml
@@ -87,10 +87,11 @@ spec:
                           type: object
                           x-kubernetes-preserve-unknown-fields: true
                         version:
-                          default: "3.2"
-                          description: Version contains the minor version for Kafka. Currently only "3.2" is supported. Leave it empty to always get the latest supported version.
+                          default: "3.4"
+                          description: Version contains the minor version for Kafka. Currently only "3.4" is supported. Leave it empty to always get the latest supported version.
                           enum:
-                            - "3.2"
+                            - "3.3"
+                            - "3.4"
                           type: string
                         zone:
                           default: ch-gva-2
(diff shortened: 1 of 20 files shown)


Kidswiss:
The end-user docs will be handled in a seperate PR

crds/exoscale.appcat.vshn.io_exoscalekafkas.yaml
```diff
@@ -87,10 +87,11 @@ spec:
                           type: object
                           x-kubernetes-preserve-unknown-fields: true
                         version:
-                          default: "3.2"
-                          description: Version contains the minor version for Kafka. Currently only "3.2" is supported. Leave it empty to always get the latest supported version.
+                          default: "3.4"
```
wejdross, line 90: out of curiosity - wouldn't that affect running instances - if any?

crds/exoscale.appcat.vshn.io_exoscalekafkas.yaml
```diff
@@ -87,10 +87,11 @@ spec:
...
+                          description: Version contains the minor version for Kafka. Currently only "3.4" is supported. Leave it empty to always get the latest supported version.
                           enum:
-                            - "3.2"
+                            - "3.3"
```
wejdross, line 93: there are 3.3 and 3.4, but in description is message, that only 3.4 is supported - does it make any sense to expose 3.3 anyway?

docs/modules/ROOT/pages/explanations/add-to-openshift-catalog.adoc
```diff
@@ -0,0 +1,98 @@
...
+[source,jsonnet]
+----
+local isOpenshift = std.startsWith(inv.parameters.facts.distribution, 'openshift');
+
+{
+  '20_xrd_vshn_postgres': xrd,
+  ...
+  [if isOpenshift then '21_openshift_template_postgresql_vshn']: osTemplate,
+}
+----
```
wejdross, line 98: love the docs :) great work

wejdross (approved):
I have just this question about Kafka which is probably unrelated to the change, other than that - accepted

crds/exoscale.appcat.vshn.io_exoscalekafkas.yaml
Kidswiss: Yes it  would, but currently there are no Kafkas on APPUiO Exoscale and at this point users can't actually provision one as 3.2 is not available anymore...

crds/exoscale.appcat.vshn.io_exoscalekafkas.yaml
Kidswiss: Yeah let's remove 3.3

PR #155, opened 2023-05-25
Adding possibility to send some alerts to customers via mailgun

wejdross:
## Summary

This adds the possibility to send alerts triggered by the created instance to the customer via mailgun.
This PR will add the required smtp secret for mailgun to the cluster, which is then used by the alertmanager config of that instance to send alerts to the configured e-mail address in the claim.
The alertmanager config itself is created using composition functions and is implemented here: https://github.com/vshn/appcat/pull/21

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/class/defaults.yml
+++ b/class/defaults.yml
@@ -28,7 +28,9 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.2.0
+        tag: v4.4.0
+
+    namespace: syn-appcat
 
     controller:
       enabled: false
@@ -36,7 +38,7 @@ parameters:
       postgres:
         enabled: false
         extraArgs: []
-        extraEnv: { }
+        extraEnv: {}
         resources:
           requests:
             cpu: 100m
@@ -100,7 +102,6 @@ parameters:
                 for: 6m
               ticket_alert: {}
 
-
     providers:
       cloudscale:
         enabled: false
@@ -151,6 +152,13 @@ parameters:
       vshn:
         enabled: false
         secretNamespace: ${crossplane:namespace}
+        emailAlerting:
+          smtpHost: "smtp.eu.mailgun.org:465"
+          smtpUsername: myuser@example.com
+          smtpPassword: "?{vaultkv:__shared__/__shared__/mailgun/smtp_password}"
+          smtpFromAddress: myuser@example.com
+          secretNamespace: syn-appcat
+          secretName: mailgun-smtp-credentials
         postgres:
           # bucket_region: 'lpg' || 'ch-gva-2'
           bucket_region: ""
@@ -160,7 +168,7 @@ parameters:
           enableNetworkPolicy: true
           secretNamespace: ${appcat:services:vshn:secretNamespace}
           # Used for deploying jobs during restores
-          controlNamespace: 'syn-appcat-control'
+          controlNamespace: "syn-appcat-control"
           sgNamespace: syn-stackgres-operator
 
           defaultPlan: standard-2
--- a/component/main.jsonnet
+++ b/component/main.jsonnet
@@ -3,8 +3,10 @@ local kap = import 'lib/kapitan.libjsonnet';
 local kube = import 'lib/kube.libjsonnet';
 local slos = import 'slos.libsonnet';
 
+
 local inv = kap.inventory();
 local params = inv.parameters.appcat;
+local pgParams = params.services.vshn.postgres;
 
 local xrdBrowseRole = kube.ClusterRole('appcat:browse') + {
   metadata+: {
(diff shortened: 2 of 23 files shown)


class/appcat.yml
```diff
@@ -2,9 +2,9 @@ parameters:
   kapitan:
     dependencies:
       - type: git
-        output_path: dependencies/appcat/manifests/${appcat:images:appcat:tag}
+        output_path: dependencies/appcat/manifests/mailgun-alerting
         source: https://github.com/vshn/appcat.git
-        ref: ${appcat:images:appcat:tag}
+        ref: mailgun-alerting
```
Kidswiss, line 7: You probably want to change this back.

class/appcat.yml
```diff
@@ -2,9 +2,9 @@ parameters:
   kapitan:
     dependencies:
       - type: git
-        output_path: dependencies/appcat/manifests/${appcat:images:appcat:tag}
+        output_path: dependencies/appcat/manifests/mailgun-alerting
```
Kidswiss, line 5: You probably want to change this back.

class/defaults.yml
```diff
@@ -28,7 +28,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.0.0
+        tag: mailgun-alerting
```
Kidswiss, line 31: You probably want to change this back.

component/main.jsonnet
```diff
@@ -66,12 +68,34 @@ local readServices = kube.ClusterRole('appcat:services:read') + {
   ],
 };
 
-{
-  '10_clusterrole_view': xrdBrowseRole,
-  [if isOpenshift then '10_clusterrole_finalizer']: finalizerRole,
-  '10_clusterrole_services_read': readServices,
+// adding namespace for syn-appcat
+local ns = kube.Namespace('syn-appcat');
+// adding missing namespace for slos
```
Kidswiss, line 73: The SLI part of the component deploys this one already. Why did you add it again here?

component/main.jsonnet
```diff
@@ -66,12 +68,34 @@ local readServices = kube.ClusterRole('appcat:services:read') + {
...
+local slos_namespace = kube.Namespace('appcat-slos');
+
+// add secret
+
+local secret = kube.Secret('mailgun-smtp-credentials') {
+  metadata+: {
+    namespace: 'syn-appcat',
+  },
+  stringData: {
+    password: pgParams.mailgun_smtp_password,
```
Kidswiss, line 83: There doesn't seem to be a vault reference in the `defaults.yaml` where does this come from?

component/main.jsonnet
```diff
@@ -66,12 +68,34 @@ local readServices = kube.ClusterRole('appcat:services:read') + {
...
+  },
+};
+
+if params.services.vshn.enabled && pgParams.enabled then
```
Kidswiss, line 87: This `if` seems to remove those files clusterrole files when generating the golden images. That will probably break a lot of stuff, please change this if statement so that it only applies to the mailgun secret.

Kidswiss (changes requested):
This removes some necessary clusterroles from the golden master, please fix that.

Also please add a summary to your PR description for the changelog.

component/main.jsonnet
wejdross: because when You setup fresh kind cluster, then namespace is never created and our makefile returns error

component/main.jsonnet
wejdross: That;s good hint, I'll move it a little bit up in hierarchy - I wanted to add valut values to syn, because each cluster has it's own secret - should it be rather here?

component/main.jsonnet
Kidswiss: What target fails for you? The `vshnpostgresql` target works for me without error.

But I would still remove it here, or we will have problems with argoCD once we deploy stuff. There should be other means to solve your issue.

component/main.jsonnet
Kidswiss: If you put the values into the vault you still need a reference in the component.

You can look at other vault values in the component to see how they are handled. For example this here: https://github.com/vshn/component-appcat/blob/master/class/defaults.yml#L125

Also add some documentation how we get those secrets, so everyone could install AppCat on a new cluster.

component/main.jsonnet
Kidswiss: This still removes some necessary files it seems.

class/defaults.yml
```diff
@@ -151,6 +151,7 @@ parameters:
       vshn:
         enabled: false
         secretNamespace: ${crossplane:namespace}
+        mailgun_smtp_password: "?{vaultkv:${cluster:tenant}/${cluster:name}/appcat/mailgun_smpt_password}"
```
Kidswiss, line 154: We should probably document where this smtp password comes from.

Also, if we only have one password for mailgun, then maybe we should not store it in a cluster specific place? Then the manual work to install appcat will be a bit less.

Kidswiss (changes requested):
Just one thing, the rest LGTM

class/defaults.yml
zugao, line 154: Is there a reason for not using camel case here?

component/main.jsonnet
```diff
@@ -66,10 +68,23 @@ local readServices = kube.ClusterRole('appcat:services:read') + {
   ],
 };
 
+// adding namespace for syn-appcat
+local ns = kube.Namespace('syn-appcat');
```
zugao, line 72: We should probably need to make it configurable in defaults like any other namespace.

zugao (commented):
Some things to clarify

class/defaults.yml
TheBigLee: Fixed

component/main.jsonnet
TheBigLee: Fixed

Kidswiss (approved):

Kidswiss (approved):

component/main.jsonnet
```diff
@@ -66,10 +68,23 @@ local readServices = kube.ClusterRole('appcat:services:read') + {
...
+local ns = kube.Namespace(params.namespace);
+
+local secret = kube.Secret(params.services.vshn.emailAlerting.secretName) {
+  metadata+: {
+    namespace: params.services.vshn.emailAlerting.secretNamespace,
+  },
+  stringData: {
+    password: params.services.vshn.emailAlerting.smtpPassword,
+  },
+};
```
glrf, line 81: Nitpick: I'd prefer if we'd only deploy the secret if we actually need it. I.e. if VSHN services are enabled.

TheBigLee: Fixed

PR #157, opened 2023-05-31
Migrating maintenance logic to comp-func

Kidswiss:
This commit removes most of the logic for handling the maintenance. It now resides in the comp-funcs.

This introduces the config key for composition functions, as we need to pass some external information into the function.

## Checklist

      changelog.
      linked in the changelog.


Kidswiss:
> P.S.: I hope we are using the config resource properly. I read a bit the documentation again and it seems ok for npow.

If they don't change it massively in the future then I don't see a problem. But on the other hand, it would not be the first Crossplane Composition Functions feature we misuse a slight bit.

PR #172, opened 2023-06-29
Add prometheus rule to aggregate AppCat billing data per cluster

Kidswiss:
## Summary

This will add a recording rule which is slightly different between appuio managed and appuio cloud. There's a new flag to handle this.
    
If the cluster is APPUiO Managed then the recording rule will get the customer's ID hardcoded into the rule.
    
If the cluster is APPUiO Cloud, then the rule will use the provided organization information.
    
Additionally, the query has been reworked so it's generic. It's now able to query all current and future AppCat services as long as the namespace where it's deployed contains a label of the pattern `label_appuio_io_billing_name=~"appcat-.+"`.

Example output of the rule with a postgres and redis instance:
Screenshot 2023-07-04 at 08 37 40

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/class/defaults.yml
+++ b/class/defaults.yml
@@ -34,6 +34,8 @@ parameters:
     namespaceLabels: {}
     namespaceAnnotations: {}
 
+    appuioManaged: true
+    tenantID: ${cluster:tenant}
 
     controller:
       enabled: false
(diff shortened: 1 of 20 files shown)


zugao (approved):
Seems good to me though I haven't tested it properly.  Many labels still do not exist so we need to make sure when we implement them they match. Also I would add an output example of a result or something that would facilitate the understanding of this prometheus rule.

Kidswiss:
I've added a screenshot with an output example for redis and postgresql.

PR #173, opened 2023-06-29
Add maintenance job

zugao:
This PR adds Redis maintenance support.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/class/defaults.yml
+++ b/class/defaults.yml
@@ -28,7 +28,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.8.0
+        tag: add-maintenance-job
 
     namespace: syn-appcat
 
--- a/class/defaults.yml
+++ b/class/defaults.yml
@@ -28,7 +28,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: add-maintenance-job
+        tag: v4.9.0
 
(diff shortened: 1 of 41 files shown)


class/defaults.yml
```diff
@@ -28,7 +28,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.8.0
+        tag: add-maintenance-job
```
zugao, line 31: This will change to v4.9.0

component/provider.jsonnet
```diff
@@ -178,6 +178,11 @@ local controllerConfigRef(config) =
           resources: [ 'schedules' ],
           verbs: [ 'get', 'list', 'watch', 'update', 'patch', 'create', 'delete' ],
         },
+        {
```
Kidswiss, line 181: I'm a bit confused why you add this rule to the provider-kubernetes? We directly apply the `Release` object without going through provider-kubernetes. So provider-kubernetes should never touch this, or am I missing something?

Kidswiss (changes requested):

component/provider.jsonnet
zugao: yeah it seems you are right I will remove this.

Kidswiss (approved):
Please don't forget to adjust the tag :)

PR #174, opened 2023-06-30
Add option to define namespace labels and annotations

glrf:
This gives us some flexibilty and can be helpful in multiple situations. The main usecase currently is that we can schedule appcat on infra nodes on OpenShift using the annotation `openshift.io/node-selector`

## Checklist

      changelog.
      linked in the changelog.


wejdross (approved):
I totally agree that it might be helpful, LGTM

PR #175, opened 2023-07-03
Fix user permission in DB namespace

glrf:
We now additionally give read access for the namespace and statefulsets. This improves the UX in the OpenShift UI as users now see the project as well as see the database deployment in the `Topology` tab.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/main.jsonnet
+++ b/component/main.jsonnet
@@ -57,7 +57,12 @@ local readServices = kube.ClusterRole('appcat:services:read') + {
   rules+: [
     {
       apiGroups: [ '' ],
-      resources: [ 'pods', 'pods/log', 'pods/status', 'events', 'services' ],
+      resources: [ 'pods', 'pods/log', 'pods/status', 'events', 'services', 'namespaces' ],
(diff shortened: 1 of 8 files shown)


component/main.jsonnet
```diff
@@ -57,7 +57,12 @@ local readServices = kube.ClusterRole('appcat:services:read') + {
   rules+: [
     {
       apiGroups: [ '' ],
-      resources: [ 'pods', 'pods/log', 'pods/status', 'events', 'services' ],
+      resources: [ 'pods', 'pods/log', 'pods/status', 'events', 'services', 'namespaces' ],
+      verbs: [ 'get', 'list', 'watch' ],
+    },
+    {
+      apiGroups: [ 'apps' ],
+      resources: [ 'statefulsets' ],
```
glrf, line 65: This is so that the database shows up in the `Topology` tab. I think this should be fine as we don't really leak more information than we do through the pod anyway

Kidswiss (approved):

PR #179, opened 2023-07-06
Add postgres SLA

zugao:
Add SLA promtheus rule for postgres

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/appcat_sla_reporter.jsonnet
+++ b/component/appcat_sla_reporter.jsonnet
@@ -1,13 +1,12 @@
+local common = import 'common.libsonnet';
 local com = import 'lib/commodore.libjsonnet';
 local kap = import 'lib/kapitan.libjsonnet';
 local kube = import 'lib/kube.libjsonnet';
-
 
 local inv = kap.inventory();
 // The hiera parameters for the component
 local params = inv.parameters.appcat;
 local sla_reporter_params = params.slos.sla_reporter;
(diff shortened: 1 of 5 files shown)


component/appcat_sla_reporter.jsonnet
```diff
@@ -76,7 +75,10 @@ local ObjectStorage = kube._Object('appcat.vshn.io/v1', 'ObjectBucket', 'appcat-
   },
 };
 
+local promRulePostgresSLA = common.PromRuleSLA(params.services.vshn.postgres.sla, 'postgres');
```
Kidswiss, line 78: I'd actually move this to `vshn_postgres.jsonnet`.

Reasoning:
We can enable and disable services individually. So it might be possible that we have a cluster where only redis is enabled, then we would not need the postgresql sla.
Also having everything that can belongs to a specific service in the same place makes it easier to handle IMHO.

Kidswiss (changes requested):
Just one thing that bothers me a bit.

component/appcat_sla_reporter.jsonnet
zugao: Actually I set it first in postgres but then I realised that maybe put them together in the same place would be easier to figure out the SLAs. But your argument is good enough and will put it back in postgres jsonnet

Kidswiss (approved):

PR #183, opened 2023-07-12
Update SGDbOps resources 

zugao:
* Update SGDBOps resource requests.
* Bump appcat version
 
## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/class/defaults.yml
+++ b/class/defaults.yml
@@ -28,7 +28,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.11.1
+        tag: v4.11.2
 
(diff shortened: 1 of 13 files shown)


Kidswiss (approved):

Kidswiss:
I'm not sure if a cluster will go into the `pendingrestart` condition if its instance profile is changed.

If so it might get restarted outside of the maintenance window.

zugao:
> I'm not sure if a cluster will go into the `pendingrestart` condition if its instance profile is changed.
> 
> If so it might get restarted outside of the maintenance window.

I tested this case and the instance is not restarted.

PR #185, opened 2023-07-12
Add missing project permissions on appcat service namespace

TheBigLee:
In order for the end user to properly see the logs in the console, the project permission is required.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/main.jsonnet
+++ b/component/main.jsonnet
@@ -70,6 +70,11 @@ local readServices = kube.ClusterRole('appcat:services:read') + {
       resources: [ 'pods/portforward' ],
       verbs: [ 'get', 'list', 'create' ],
     },
+    {
+      apiGroups: [ '', 'project.openshift.io' ],
+      resources: [ 'projects' ],
+      verbs: [ 'get' ],
+    },
   ],
 };
 
--- a/tests/golden/apiserver/appcat/appcat/10_clusterrole_services_read.yaml
+++ b/tests/golden/apiserver/appcat/appcat/10_clusterrole_services_read.yaml
@@ -35,3 +35,10 @@ rules:
       - get
       - list
       - create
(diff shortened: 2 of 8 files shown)


component/main.jsonnet
```diff
@@ -70,6 +70,11 @@ local readServices = kube.ClusterRole('appcat:services:read') + {
       resources: [ 'pods/portforward' ],
       verbs: [ 'get', 'list', 'create' ],
     },
+    {
+      apiGroups: [ '""', 'project.openshift.io' ],
```
glrf, line 74: `'""'` seems wrong/unnecessary. Why did you add that?

apis/generated/appcat.vshn.io_objectbuckets.yaml
glrf, line 1: I don't think we commit the `apis` folder anymore. We should probably add that to the gitignore

glrf (approved):
TIL

Please fix the golden tests and remove the `apis` directory. Otherwise looks good to me

component/main.jsonnet
TheBigLee: It is specified like this in the default `view` clusterRole. Not sure why though

apis/generated/appcat.vshn.io_objectbuckets.yaml
TheBigLee: Removed them again

component/main.jsonnet
glrf: Hmm interesting. Maybe some OpenShift weirdness/backwards compatibility thing?

But also to get the same output you'll need to use 

```suggestion
      apiGroups: [ '', 'project.openshift.io' ],
```

simu (approved):

PR #187, opened 2023-07-18
Add SA, Clusterrole and Clusterrolebinding for redis restores

TheBigLee:
This PR adds the necessary SA, Clusterrole and Clusterrolebinding in order to provide the necessary permissions for restoring a redis instance.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/provider.jsonnet
+++ b/component/provider.jsonnet
@@ -134,6 +134,16 @@ local controllerConfigRef(config) =
           verbs: [ 'get', 'list', 'watch', 'create', 'watch', 'patch', 'update', 'delete' ],
         },
         {
+          apiGroups: [ 'apps' ],
+          resources: [ 'statefulsets/scale' ],
+          verbs: [ 'update', 'patch' ],
+        },
+        {
+          apiGroups: [ 'apps' ],
+          resources: [ 'statefulsets' ],
+          verbs: [ 'get' ],
+        },
+        {
           apiGroups: [ 'rbac.authorization.k8s.io' ],
           resources: [ 'clusterroles' ],
           resourceNames: [ 'appcat:services:read' ],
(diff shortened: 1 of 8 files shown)


zugao (commented):
The maintenance RBAC are in `vshn_services.jsonnet`. You might want to consider to put them together in the same place. Also please consider using the same RBAC rules as for maintenance. @Kidswiss what do you think about that? Should we separate? I feel we are creating too much entropy here.

Kidswiss:
@zugao Moving them to the same place makes sense. But re-using the same RBAC rules for maintenance and backup/restore to me feels like a bad idea. Maintenance probably needs different RBAC than the restore.

TheBigLee:
What about the RBAC rules for postgresql? They are also in vshn_postgres.jsonnet and use different rules than what we need for redis. How should we handle that?

zugao:
The reason as to why I created another file is that the maintenance RBAC rules can be reused for other helm based services. If you believe that restore RBAC also can be reused by other Helm based services that you should move them there otherwise keep in redis file. Postgres is not a Helm based service therefore in postgres file.

TheBigLee:
> The reason as to why I created another file is that the maintenance RBAC rules can be reused for other helm based services. If you believe that restore RBAC also can be reused by other Helm based services that you should move them there otherwise keep in redis file. Postgres is not a Helm based service therefore in postgres file.

Some of them might be reusable but probably not all. It really depends on the helm chart (eg. if it uses statefulsets or deployments)
I fear that having a global set of RBAC for all kind of (helm) services might end up in very wide set of rbac rules for that service account with way too many permissions.

Kidswiss:
> > The reason as to why I created another file is that the maintenance RBAC rules can be reused for other helm based services. If you believe that restore RBAC also can be reused by other Helm based services that you should move them there otherwise keep in redis file. Postgres is not a Helm based service therefore in postgres file.
> 
> Some of them might be reusable but probably not all. It really depends on the helm chart (eg. if it uses statefulsets or deployments) I fear that having a global set of RBAC for all kind of (helm) services might end up in very wide set of rbac rules for that service account with way too many permissions.

In that case I propose to leave it as is.

Kidswiss (approved):

zugao (approved):
LGTM

PR #191, opened 2023-07-26
Adding necessary field names to handle external IPs in connection secret 

wejdross:
This PR is strictly connected with: https://github.com/vshn/appcat/pull/46 
I need to add fields to connection secret so crossplane can write to them

## Checklist

      changelog.
      linked in the changelog.


component/vshn_postgres.jsonnet
```diff
@@ -37,6 +37,8 @@ local connectionSecretKeys = [
   'POSTGRESQL_PORT',
   'POSTGRESQL_USER',
   'POSTGRESQL_PASSWORD',
+  'PRIMARY_IP',
```
Kidswiss, line 40: Maybe call it `EXTERNAL_IP` or `LOADBALANCER_IP`, that would make it clearer.

component/vshn_postgres.jsonnet
```diff
@@ -37,6 +37,8 @@ local connectionSecretKeys = [
...
+  'REPLICAS_IP',
```
Kidswiss, line 41: I wouldn't expose the replica IPs, we also hide them internally.

Kidswiss (changes requested):
Well that looks easy :D

Just two remarks about naming and the replica ips

component/vshn_postgres.jsonnet
wejdross: I like LOADBALANCER_IP I'll switch name to that

component/vshn_postgres.jsonnet
wejdross: I think that replicas could be very important for our customers, You normally don't want to do analytics using master node/instance, this is why I decided to expose REPLICA_IP, also it might be important for customers to make fast snapshot of database for local development purposes - also replica is much better choice.

What are counter arguments?

component/vshn_postgres.jsonnet
```diff
@@ -403,6 +413,8 @@ local sgCluster = {
 
     comp.FromCompositeFieldPath('spec.parameters.backup.schedule', 'spec.forProvider.manifest.spec.configurations.backups[0].cronSchedule'),
     comp.FromCompositeFieldPath('spec.parameters.backup.retention', 'spec.forProvider.manifest.spec.configurations.backups[0].retention'),
+    comp.FromCompositeFieldPath('spec.parameters.network.serviceType', 'spec.forProvider.manifest.spec.postgresServices.primary.type'),
```
Kidswiss, line 416: I just saw, that you also have logic in the composition function. Please do these patches there as well. It's confusing if the logic is spread over the composition function and the legacy P+T compostion.

tests/golden/vshn/appcat/appcat/21_composition_vshn_postgres.yaml
```diff
@@ -621,6 +632,12 @@ spec:
         - fromConnectionSecretKey: POSTGRESQL_PASSWORD
           name: POSTGRESQL_PASSWORD
           type: FromConnectionSecretKey
+        - fromConnectionSecretKey: PRIMARY_IP
```
Kidswiss, line 635: Is this actually necessary? You already export the values as a connectiondetail in the compostion function.

tests/vshn.yml
```diff
@@ -38,6 +38,7 @@ parameters:
     services:
       vshn:
         enabled: true
+        externalDatabaseConnectionsEnabled: true
```
Kidswiss, line 41: Can you add this parameter to the default as well? But set it to false, as we can only use it on specific clusters.

Otherwise the compilation will probably fail for the remaining clusters, as there is no field with that name available.

Kidswiss (changes requested):

Kidswiss (approved):

PR #203, opened 2023-08-21
Add webhook support for the controller

Kidswiss:
The controller has also been renamed, as it now doesn't only handle postgres anymore, but also redis webhooks as well.
    
It has also been moved to the `syn-appcat` namespace so we can get rid of `syn-appcat-controllers` namespace.

## Checklist

      changelog.
      linked in the changelog.


class/defaults.yml
```diff
@@ -32,7 +32,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.19.0
+        tag: add/postgresql-webhooks
```
Kidswiss, line 35: Wil be adjusted before release.

zugao (approved):
LGTM, I suppose you tested it successfully in Lab cluster

PR #224, opened 2023-09-13
Add disable slos alerts parameter

zugao:
This small PR adds the possibility to disable alerts when SLOS is still enabled. This parameters helps in dev environment.

## Checklist

      changelog.
      linked in the changelog.


component/slos.libsonnet
```diff
@@ -7,7 +7,7 @@ local kube = import 'lib/kube.libjsonnet';
 local inv = kap.inventory();
 // The hiera parameters for the component
 local params = inv.parameters.appcat;
-
+local alertsDisabled = if inv.parameters.appcat.slos.alertsEnabled then false else true;
```
wejdross, line 10: please rename this variable to alertsEnabled so we have some order :D

wejdross (changes requested):

component/slos.libsonnet
zugao: yeah I guess I exagerated here :D

zugao: Actually I simplified the variable, now should be good

wejdross: looks better, but naming is still weird, in class You use alertsEnabled in jsonnet alertsDisabled and You inverse it - why?

wejdross: giving it 3rd look - why do You even declare new variable and just not use param directly?

component/slos.libsonnet
```diff
@@ -46,6 +47,7 @@ local prometheusRule(name) =
         } for r in g.rules ],
       }
       for g in super.groups
+      if !(alertsDisabled && std.startsWith(g.name, 'sloth-slo-alerts'))
```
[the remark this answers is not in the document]
wejdross: such thing is much clearer for me and it doesn't introduce superfluous variables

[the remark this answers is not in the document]
zugao: this will not even compile and you need to add `== false` that's why I came up with alertsDisabled. It's easier to read the already complicated if statement.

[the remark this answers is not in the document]
wejdross: I don't like this solution, but I trust You picked the best option

wejdross (approved):

PR #229, opened 2023-09-20
Generate a valid `kustomization.yaml` when SLOs are disabled

simu:
We had various reports where catalog compilation sometimes failed with `Error: kustomization.yaml is empty`. After some investigation, it turns out that this error is caused by kustomize v5.1.1 when a completely empty file is provided as `kustomization.yaml`.

We fix the issue by generating a `kustomization.yaml` which contains `resources: []` when SLOs are disabled.

## Checklist

      changelog.
      linked in the changelog.


Kidswiss (approved):
LGTM

Thanks for investigating!

PR #234, opened 2023-09-27
Patch bucketdeletionpolicy

Kidswiss:
Patches the new bucketDeletionPolicy from the claim to the various bucket objects.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/class/defaults.yml
+++ b/class/defaults.yml
@@ -39,7 +39,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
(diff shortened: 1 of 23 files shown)


zugao (approved):
I presume it was hardcoded from the composition until now?

Kidswiss:
> I presume it was hardcoded from the composition until now?

Yes, there was a config key in the defaults.yaml which controlled the value. That value is now just the default in the XRD. So if nothing is specified in the claim then there will be no change.

PR #235, opened 2023-09-28
Enable resource limits for sidecars and set sensible values

TheBigLee:
We need to set the resource limits on the sidecars as well and then set some sensible values for them.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/class/defaults.yml
+++ b/component/class/defaults.yml
@@ -275,103 +275,103 @@ parameters:
           sideCars:
             createBackup:
               requests:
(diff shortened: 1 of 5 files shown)


zugao (approved):
Looks good so far. In case cluster-controller does not leak memory anymore it should be fine otherwise we might increase it to 1,5GB so that the pod is not killed before maintenance.

zugao:
I increase the memory to 2Gi for cluster-controller container.

PR #238, opened 2023-10-04
Refactor component and add appcat package 

TheBigLee:
This PR will move the component to the `component` folder and add a package in `package` to deply appcat together with crossplane.

## Checklist

      changelog.
      linked in the changelog.


wejdross:
Sorry Bigli, but I'm unable to review this as I have no idea about project syn

zugao:
> Sorry Bigli, but I'm unable to review this as I have no idea about project syn

I don't want to sound harsh but you will need to learn a thing or two since a part of our job is to work with project syn. We are currently low on manpower I would encourage you to take a bit of time and effort and learn some basic stuff. This also applies to me and other team members.

wejdross:
> > Sorry Bigli, but I'm unable to review this as I have no idea about project syn
> 
> I don't want to sound harsh but you will need to learn a thing or two since a part of our job is to work with project syn. We are currently low on manpower I would encourage you to take a bit of time and effort and learn some basic stuff. This also applies to me and other team members.

yeah, sure, but currently I don't have time and resources to do that. This is why I prefer to tell it in the beginning that I'm not right person to do the review. If I read docs/tutorial how this software work, I won't be able to do the review as well. Maybe someone from Aldebaran could do the review if we're unable.

zugao (commented):
Seems ok but there are too many files changed. There's some thing about lab config that I believe should not be there.

package/lab.yaml
```diff
@@ -0,0 +1,8 @@
+# Include this class only on test clusters
+classes:
```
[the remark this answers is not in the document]
TheBigLee: The idea here is, that appuio lab will use that configuration (and hence default to master branch) instead of the latest tag)

The version of the component however, can still be overridden in the tenant configuration (e.g to a branch to test a feature)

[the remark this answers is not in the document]
zugao: As discussed we will remove the `lab.yaml` file and will just move the config to the `syn-tenant-repo`

zugao (approved):
LGTM

package/main.yaml
```diff
@@ -0,0 +1,66 @@
+applications:
+  - appcat
+  - crossplane
+parameters:
+  pkg.appcat:
+    componentVersion: package_appcat
```
simu, line 6: You'll probably want to adjust this to something that isn't a feature branch name before merging

.github/workflows/release.yaml
```diff
@@ -30,3 +30,19 @@ jobs:
           # Ensure target branch for release is "master"
           commit: master
           token: ${{ secrets.GITHUB_TOKEN }}
+  update-version:
```
simu, line 33: I'm not sure this will work as you'd like it to, unless you plan to always use package version `master`.

simu (commented):
Package config LGTM overall. See inline comments for things that I might do differently.

package/main.yaml
TheBigLee: The problem here is a classical chicken egg problem. If I do, the golden tests will fail.
However, there is a github action that will update this automatically when doing a release.

.github/workflows/release.yaml
TheBigLee: We do plan to only use `master` for the package

PR #239, opened 2023-10-09
adding VSHNRedis ServiceMonitor to expose Redis metrics in Openshift

wejdross:
This is a simple change which:
- adds Redis sidecar with metrics exporter
- adds ServiceMonitor so metrics are available within namespace in Openshift console
- disables securityContext in metrics exporter, so default from Openshift is used

Tested with success on LAB cluster

## Checklist

      changelog.
      linked in the changelog.


component/component/vshn_redis.jsonnet
```diff
@@ -93,6 +93,25 @@ local restoreRole = kube.ClusterRole(restoreRoleName) {
     },
   ],
 };
+local helmMonitoringClusterRoleName = 'allow-helm-monitoring-resources';
+local helmMonitoringClusterRole = kube.ClusterRole(helmMonitoringClusterRoleName) {
```
zugao, line 97: You can remove the variable defined above and just pass the name here.

component/component/vshn_redis.jsonnet
```diff
@@ -377,6 +396,16 @@ local composition =
...
           },
           values: {
+            metrics: {
+              enabled: true,
+              containerSecurityContext: {
+                enabled: false,
+              },
+              serviceMonitor: {
+                enabled: true,
+                namespace: '',  // patched
```
zugao, line 406: Where this is being patched?

component/component/vshn_redis.jsonnet
```diff
@@ -377,6 +396,16 @@ local composition =
             version: redisParams.helmChartVersion,
           },
           values: {
+            metrics: {
+              enabled: true,
+              containerSecurityContext: {
+                enabled: false,
```
zugao, line 402: This I suppose is to make it work on OpenShift?

zugao (commented):
Looks good overall but there are things to take a look and change and please remove `compiled` `dependencies` and `vendor` directories from this PR.

component/component/vshn_redis.jsonnet
wejdross: in line 645: 
```
comp.FromCompositeFieldPathWithTransformPrefix('metadata.labels[crossplane.io/composite]', 'spec.forProvider.values.metrics.serviceMonitor.namespace', 'vshn-redis'),
```

wejdross: in line 645: 
```
comp.FromCompositeFieldPathWithTransformPrefix('metadata.labels[crossplane.io/composite]', 'spec.forProvider.values.metrics.serviceMonitor.namespace', 'vshn-redis'),
```

component/component/vshn_redis.jsonnet
wejdross: exactly

zugao: Don't we have a variable to figure out if the cluster has distribution OpenShift? We want our AppCat to run also on on OpenShift clusters. Like [here](https://github.com/vshn/component-appcat/blob/master/component/component/vshn_redis.jsonnet#L37).

component/component/vshn_redis.jsonnet
```diff
@@ -94,6 +94,25 @@ local restoreRole = kube.ClusterRole(restoreRoleName) {
   ],
 };
 
+local helmMonitoringClusterRole = kube.ClusterRole('allow-helm-monitoring-resources') {
+  rules: [
+    {
+      apiGroups: [ 'monitoring.coreos.com' ],
+      resources: [ 'servicemonitors' ],
+      verbs: [ '*' ],
```
zugao, line 102: I was thinking about this configuration, are we sure we need all verbs? Maybe we need only `list, watch and get` only.

wejdross: it also requires create, update and delete, therefore I used `'*'` asteriks value - easier to read :)

PR #247, opened 2023-10-17
add liveness and readiness probes to appcat apiserver

wejdross:
Due to timeout issues with apiserver as mitigation of an issue I'm introducing liveness and readiness probes which should restart Apiserver whenever it fails to respond. If that won't be enough I suggest adding another probe based on shell command. 

Tested it locally and on lab - it works, it checks.


component/component/appcat_apiserver.jsonnet
```diff
@@ -92,6 +92,28 @@ local apiserver = loadManifest('aggregated-apiserver.yaml') {
...
+                  scheme: 'HTTPS',
+                },
+                timeoutSeconds: 2,
+                successThreshold: 1,
+                initialDelaySeconds: 10,
+                failureThreshold: 3,
+              },
+              readinessProbe: {
+                httpGet: {
+                  path: '/livez',
```
zugao, line 108:
```suggestion
                  path: '/readyz',
```

zugao (changes requested):
Have you tried this in lab? Not sure if the probe bypasses tls with `--insecure`

zugao (approved):
LGTM

PR #251, opened 2023-10-27
Fix restore operation not working for PostgreSQL by VSHN

zugao:
Restore operation does not work anymore for PostgreSQL by VSHN because in some of out composition functions we are referencing name `cluster` for and SGCluster object in the _iof_ function. We need to keep the same name between restore and main compositions. This name change is purely internal and is not used outside the code.
Tested it in local cluster and it worked.

## Checklist

      changelog.
      linked in the changelog.


wejdross (approved):
yeah, such a small thing :D

PR #256, opened 2023-11-03
Add/non slo redis alerts

wejdross:
Enabling Redis System Metrics, because of:
- `REDIS_EXPORTER_INCL_SYSTEM_METRICS | Whether to include system metrics like total_system_memory_bytes, defaults to false.`
- all rules definitions around Internet use them, so I prefer to reuse them
- I made nice workaround to those metrics, but it was much longer with bunch of regexes and I dropped that afterwards

Redis Storage was simply copied from PostgreSQL storage

### DIFF
--- a/component/component/common.libsonnet
+++ b/component/component/common.libsonnet
@@ -203,23 +203,23 @@ local generatePrometheusNonSLORules(serviceName, memoryContainerName, additional
           metadata: {
             apiVersion: 'monitoring.coreos.com/v1',
             kind: 'PrometheusRule',
-            name: '%s-rules' % serviceName,
+            name: '%s-rules' % serviceNameLower,
           },
           spec: {
             groups: [
               {
-                name: '%s-general-alerts' % serviceName,
+                name: '%s-general-alerts' % serviceNameLower,
                 rules: [
 
                   {
-                    name: '%s-storage' % serviceName,
+                    name: '%s-storage' % serviceNameLower,
                     alert: serviceName + 'PersistentVolumeFillingUp',
                     annotations: {
                       description: 'The volume claimed by the instance {{ $labels.name }} in namespace {{ $labels.label_appcat_vshn_io_claim_namespace }} is only {{ $value | humanizePercentage }} free.',
                       runbook_url: 'https://runbooks.prometheus-operator.dev/runbooks/kubernetes/kubepersistentvolumefillingup',
                       summary: 'PersistentVolume is filling up.',
                     },
-                    expr: std.strReplace(bottomPod('%(availablePercent)s < 0.03 and %(usedStorage)s > 0 %(unlessExcluded)s' % queries), toReplace, 'vshn-' + std.asciiLower(serviceName)),
+                    expr: std.strReplace(bottomPod('%(availablePercent)s < 0.03 and %(usedStorage)s > 0 %(unlessExcluded)s' % queries), toReplace, 'vshn-' + serviceNameLower),
                     'for': '1m',
                     labels: {
                       severity: 'critical',
@@ -227,14 +227,13 @@ local generatePrometheusNonSLORules(serviceName, memoryContainerName, additional
                     },
                   },
                   {
-                    name: '%s-storage' % serviceName,
                     alert: serviceName + 'PersistentVolumeFillingUp',
                     annotations: {
                       description: 'Based on recent sampling, the volume claimed by the instance {{ $labels.name }} in namespace {{ $labels.label_appcat_vshn_io_claim_namespace }} is expected to fill up within four days. Currently {{ $value | humanizePercentage }} is available.',
                       runbook_url: 'https://runbooks.prometheus-operator.dev/runbooks/kubernetes/kubepersistentvolumefillingup',
                       summary: 'PersistentVolume is filling up.',
                     },
-                    expr: std.strReplace(bottomPod('%(availablePercent)s < 0.15 and %(usedStorage)s > 0 and predict_linear(%(availableStorage)s[6h], 4 * 24 * 3600) < 0  %(unlessExcluded)s' % queries), toReplace, 'vshn-' + std.asciiLower(serviceName)),
+                    expr: std.strReplace(bottomPod('%(availablePercent)s < 0.15 and %(usedStorage)s > 0 and predict_linear(%(availableStorage)s[6h], 4 * 24 * 3600) < 0  %(unlessExcluded)s' % queries), toReplace, 'vshn-' + serviceNameLower),
                     'for': '1h',
                     labels: {
                       severity: 'warning',
@@ -248,7 +247,7 @@ local generatePrometheusNonSLORules(serviceName, memoryContainerName, additional
                       // runbook_url: 'TBD',
                       summary: 'Memory usage critical',
                     },
-                    expr: std.strReplace(topPod('(container_memory_working_set_bytes{container="%s"}  / on(container,pod,namespace)  kube_pod_container_resource_limits{resource="memory"} * 100) > 90') % memoryContainerName, toReplace, 'vshn-' + std.asciiLower(serviceName)),
+                    expr: std.strReplace(topPod('(container_memory_working_set_bytes{container="%s"}  / on(container,pod,namespace)  kube_pod_container_resource_limits{resource="memory"} * 100) > 90') % memoryContainerName, toReplace, 'vshn-' + serviceNameLower),
                     'for': '120m',
                     labels: {
                       severity: 'warning',
--- a/component/component/common.libsonnet
+++ b/component/component/common.libsonnet
@@ -134,31 +134,6 @@ local getAppCatImageString() = params.images.appcat.registry + '/' + params.imag
 
 local getApiserverImageString() = params.images.apiserver.registry + '/' + params.images.apiserver.repository + ':' + getApiserverImageTag();
 
-local promRuleSLA(value, service) = kube._Object('monitoring.coreos.com/v1', 'PrometheusRule', 'vshn-' + std.asciiLower(service) + '-sla') {
-  metadata+: {
-    labels: {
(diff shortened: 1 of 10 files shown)


component/component/vshn_redis.jsonnet
```diff
@@ -384,6 +384,82 @@ local composition =
...
+                               manifest+: {
+                                 metadata: {
+                                   name: 'redis-rules',
+                                 },
+                                 local bottomPod(query) = 'label_replace( bottomk(1, %s) * on(namespace) group_left(label_appcat_vshn_io_claim_namespace) kube_namespace_labels, "name", "$1", "namespace", "vshn-redis-(.+)-.+")' % query,
+                                 local topPod(query) = 'label_replace( topk(1, %s) * on(namespace) group_left(label_appcat_vshn_io_claim_namespace) kube_namespace_labels, "name", "$1", "namespace", "vshn-redis-(.+)-.+")' % query,
+                                 spec: {
+                                   groups: [
+                                     {
+                                       name: 'redis-memory',
```
TheBigLee, line 400: This looks wrong. Why is the top group called `redis-memory` but contains all kind of alerts?

wejdross: renamed it with `name: 'redis-non-slo-alerts',` that was a leftover after development, when I started with memory first

Kidswiss:
> all rules definitions around Internet use them, so I prefer to reuse them

I get why this could be a reason to go with the specific metrics but here's an idea:

If we use the kube-state-metrics like with PostgreSQL for memory and storage then we can put the generation of those rules into jsonnet functions and just parameterize them. Like what I did for the sloth rules.

Then we won't run into the risk of copy/paste errors, and we know all the metrics are from the same source. If we ever need to adjust them, we can do it in one place.

wejdross:
Ok, we could abstract that, but we still need to enable additional metrics for both customers and us, especially if we hit clusters monitoring, I'm flexible with both solutions - @TheBigLee be the tiebreaker :D 

Those are rules I was talking about, there are few pages aggregating Prometheus rules and they all reuse those additional metrics, this is why I think it's worth enabling them
https://samber.github.io/awesome-prometheus-alerts/rules.html#redis

Kidswiss:
Sure, I'm not against enabling the additional metrics, they will be useful for additional alerts and dashboards.

I'm just saying we should use the kube-state-metrics for storage and memory, as they are service agnostic.

component/tests/golden/vshn/appcat/appcat/21_composition_vshn_postgresrestore.yaml
```diff
@@ -1103,83 +1099,23 @@ spec:
                         for: 1h
                         labels:
                           severity: warning
-                  - name: postgresql-memory
```
zugao, line 1106: Can you make sure nothing is changed for PostgreSQL? You introduced a generic function but I expect no changes for existing PostgreSQL code.

wejdross: I just made it the closest possible to original

component/component/common.libsonnet
```diff
@@ -173,6 +173,88 @@ local argoCDAnnotations() = {
...
 
+local bottomPod(query) = 'label_replace( bottomk(1, %(query)s) * on(namespace) group_left(label_appcat_vshn_io_claim_namespace) kube_namespace_labels, "name", "$1", "namespace", "vshn-replacemeplease-(.+)-.+")' % query;
+local topPod(query) = 'label_replace( topk(1, %(query)s) * on(namespace) group_left(label_appcat_vshn_io_claim_namespace) kube_namespace_labels, "name", "$1", "namespace", "vshn-replacemeplease-(.+)-.+")' % query;
+
+local generatePrometheusNonSLORules(serviceName, additionalAlerts) = {
+  // standardized lowercase regardless of what came as input
+  local serviceNameLower = std.asciiLower(serviceName),
+  local toReplace = 'vshn-replacemeplease',
+
+  local memoryContainerName = if serviceName == 'redis' then 'redis' else 'patroni',
```
Kidswiss, line 184: Why not make `memoryContainerName` another argument for the function? This if cascade won't scale.

component/component/vshn_postgres.jsonnet
```diff
@@ -762,159 +762,90 @@ local clusterRestoreConfig = {
   ],
 };
 
-local prometheusRule = {
-  name: 'prometheusrule',
-  base: comp.KubeObject('monitoring.coreos.com/v1', 'PrometheusRule') + {
```
Kidswiss, line 767: Don't drop `name` and `base`, they are required fields for the composition.

Kidswiss (changes requested):
The rendered composition looks a bit broken. Also, before merging this test it on the lab and also check if the existing dashboards on insights need adjustments.

wejdross:
@Kidswiss  @zugao 
MR is ready to be roasted :D

component/component/common.libsonnet
```diff
@@ -173,6 +173,97 @@ local argoCDAnnotations() = {
...
+    },
+    spec: {
+      providerConfigRef: {
+        name: 'kubernetes',
+      },
+      forProvider+: {
+        manifest+: {
+          metadata: {
+            apiVersion: 'monitoring.coreos.com/v1',
+            kind: 'PrometheusRule',
```
TheBigLee, line 205:
```suggestion
          apiVersion: 'monitoring.coreos.com/v1',
          kind: 'PrometheusRule',
          metadata: {
```

Afaik `apiVersion` and `kind` need to be on the same level as `metadata` and not inside of it.

component/tests/golden/vshn/appcat/appcat/21_composition_vshn_postgres.yaml
```diff
@@ -1001,22 +1003,21 @@ spec:
                         for: 1h
                         labels:
                           severity: warning
-                  - name: postgresql-memory
-                    rules:
```
Kidswiss, line 1005: This looks broken, `name` is now deeper and the `rules` field is missing.

component/tests/golden/vshn/appcat/appcat/21_composition_vshn_postgres.yaml
```diff
@@ -1001,22 +1003,21 @@ spec:
...
                       - alert: PostgreSQLMemoryCritical
                         annotations:
                           description: |-
-                            The memory claimed by the instance {{ $labels.name }} in namespace {{ $labels.label_appcat_vshn_io_claim_namespace }} has been over 85% for 2 hours.
```
Kidswiss, line 1009: Why did the value change from 85% to 90%?

component/tests/golden/vshn/appcat/appcat/21_composition_vshn_postgres.yaml
```diff
@@ -1001,22 +1003,21 @@ spec:
...
+                            The memory claimed by the instance {{ $labels.name }} in namespace {{ $labels.label_appcat_vshn_io_claim_namespace }} has been over 90% for 2 hours.
                               Please reducde the load of this instance, or increase the memory.
                           summary: Memory usage critical
                         expr: label_replace( topk(1, (container_memory_working_set_bytes{container="patroni"}  /
                           on(container,pod,namespace)  kube_pod_container_resource_limits{resource="memory"}
-                          * 100) > 85) * on(namespace) group_left(label_appcat_vshn_io_claim_namespace)
```
Kidswiss, line 1014: Here as well.

component/tests/golden/vshn/appcat/appcat/21_composition_vshn_postgres.yaml
```diff
@@ -1001,22 +1003,21 @@ spec:
...
+                          * 100) > 90) * on(namespace) group_left(label_appcat_vshn_io_claim_namespace)
                           kube_namespace_labels, "name", "$1", "namespace", "vshn-postgresql-(.+)-.+")
                         for: 120m
                         labels:
                           severity: critical
                           syn_team: schedar
+                        name: postgresql-memory
```
Kidswiss, line 1020: Here's the `name` field that should be higher up in the structure.

Kidswiss (changes requested):
Some things look a bit broken in the rendered yaml files.

Also please test it on the lab to see if something in the dashboards would also need fixing.

zugao (approved):
As Simon already pointed out there are changes in postgres files which should not be there. This PR should not change anything in Postgres, at best only the names.

zugao (changes requested):
Changes are requested

component/tests/golden/vshn/appcat/appcat/21_composition_vshn_postgres.yaml
TheBigLee: I changed it back to 85%

component/tests/golden/vshn/appcat/appcat/21_composition_vshn_postgres.yaml
TheBigLee: Fixed

component/tests/golden/vshn/appcat/appcat/21_composition_vshn_postgres.yaml
TheBigLee: Fixed

TheBigLee:
> As Simon already pointed out there are changes in postgres files which should not be there. This PR should not change anything in Postgres, at best only the names.

This has been fixed now

TheBigLee:
> Also please test it on the lab to see if something in the dashboards would also need fixing.

What do you mean? The dashboards are not affected by this, as the dashboards use the metrics and not the alertRules. And nothing changed with the metrics themselves.

zugao (approved):
LGTM, I haven't tested it.

Kidswiss (approved):

PR #257, opened 2023-11-07
add mailgun and alerting alerting to Redis

wejdross:
Simply adding mailgun alerting and custom Alertmanager configs to Redis

## Checklist

      changelog.
      linked in the changelog.


component/component/vshn_redis.jsonnet
```diff
@@ -509,6 +509,12 @@ local composition =
                 controlNamespace: params.services.controlNamespace,
                 restoreSA: 'redisrestoreserviceaccount',
                 quotasEnabled: std.toString(params.services.vshn.quotasEnabled),
+                emailAlertingEnabled: std.toString(params.services.vshn.emailAlerting.enabled),
```
Kidswiss, line 512: I know I start to sound like a broken record in all my reviews, but maybe a function to generate these configs will avoid copy/pasting in the future.

Kidswiss (changes requested):

component/class/defaults.yml
```diff
@@ -54,7 +54,8 @@ parameters:
     appuioManaged: true
     tenantID: ${cluster:tenant}
     quotasEnabled: false
-    grpcEndpoint: unix-abstract:crossplane/fn/default.sock
+    grpcEndpoint: 172.17.0.1:9547
```
TheBigLee, line 57: this should be reverted before merging

component/class/defaults.yml
```diff
@@ -39,7 +39,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.41.1
+        tag: redis_user_alerting
```
zugao, line 42: Remember to change that

component/class/defaults.yml
```diff
@@ -54,6 +54,7 @@ parameters:
     appuioManaged: true
     tenantID: ${cluster:tenant}
     quotasEnabled: false
+    #grpcEndpoint: 172.17.0.1:9547
```
zugao, line 57: Remove this

zugao (approved):
Just some small changes otherwise LGTM

component/class/defaults.yml
TheBigLee: fixed

component/class/defaults.yml
TheBigLee: fixed

component/component/vshn_redis.jsonnet
TheBigLee: Can you clarify what exactly you mean? How does a function help here? It's only passing a parameter to the configMap for the xfn container

Kidswiss: We have the same `emailAlerting*` lines repeated over and over for each service. Each time copy pasted, this is bad practice for various reasons:

* Copy paste errors happen
* We would need to change all files again if there are changes

If we instead just have a jsonnet function where we pass `params.services.vshn.emailAlerting` and it returns all those lines, we would make it much less error prone to add a new service with email alerting. The we just need to call the function here and we're done. If there are any changes in the future we'd only have to adjust the function and all services will get the changes.

TheBigLee: Ah I understand now. Refactored the code accordingly. Please check again.

Kidswiss (approved):

PR #258, opened 2023-11-07
Forward events from redis objects to the redis claim of that instance

TheBigLee:
We want to forward the events of the redis objects (eg. statefulset) to the claim so that the end-user can see the events in the claim namespace.

We leverage our controller's functionality by setting the `vshn.net/forward-events-to` to the relevant objects.

## Checklist

      changelog.
      linked in the changelog.


Kidswiss (commented):
LGTM

Kidswiss (approved):

package/main.yaml
```diff
@@ -3,11 +3,11 @@ applications:
   - crossplane
 parameters:
   pkg.appcat:
-    componentVersion: v2.6.0
+    componentVersion: feature/sts_error_prop
     image:
       registry: ghcr.io
       repository: vshn/appcat
-      tag: v4.40.0
+      tag: feature/sts_error_prop
```
zugao, line 10: Remember to update the tag

package/main.yaml
```diff
@@ -3,11 +3,11 @@ applications:
   - crossplane
 parameters:
   pkg.appcat:
-    componentVersion: v2.6.0
+    componentVersion: feature/sts_error_prop
```
zugao, line 6: Remember to update the tag

zugao (approved):
LGTM but we might already create a generic jsonnet function that would enable event forwarding for any service.

package/main.yaml
TheBigLee: fixed

package/main.yaml
TheBigLee: fixed

TheBigLee:
> LGTM but we might already create a generic jsonnet function that would enable event forwarding for any service.

No I wouldn't do that. As we will use FunctionIO for all future services (and will have to move redis to full functionIO as well as some point), it's better to create a generic function there when the time comes.

PR #262, opened 2023-11-21
Make Crossplane 1.14 compatible

Kidswiss:
For compositions where we use functions, they had to be re-structured, as in such cases the P+T part is just treated as another function.
    
For all non-function compositions no changes are necessary.

## Checklist

      changelog.
      linked in the changelog.


component/class/defaults.yml
```diff
@@ -54,7 +62,7 @@ parameters:
     appuioManaged: true
     tenantID: ${cluster:tenant}
     quotasEnabled: false
-    grpcEndpoint: unix-abstract:crossplane/fn/default.sock
+    grpcEndpoint: host.docker.internal:9443
```
TheBigLee, line 65: This is probably wrong

component/component/provider.jsonnet
```diff
@@ -248,7 +255,6 @@ local controllerConfigRef(config) =
           }
         ),
       },
-      sa,
```
TheBigLee, line 251: Why is the SA object removed here?
Furthermore the SA that is not applied anymore is still referenced in the `rolebinding` object

component/component/vshn_postgres.jsonnet
```diff
@@ -630,7 +628,6 @@ local sgObjectStorage = {
     comp.FromCompositeFieldPathWithTransformPrefix('metadata.labels[crossplane.io/composite]', 'spec.forProvider.manifest.metadata.name', 'sgbackup'),
     comp.FromCompositeFieldPathWithTransformPrefix('metadata.labels[crossplane.io/composite]', 'spec.forProvider.manifest.metadata.namespace', 'vshn-postgresql'),
     comp.FromCompositeFieldPath('metadata.labels[crossplane.io/composite]', 'spec.forProvider.manifest.spec.s3Compatible.bucket'),
-    comp.FromCompositeFieldPath('metadata.labels[crossplane.io/claim-namespace]', 'spec.forProvider.spec.writeConnectionSecretToRef.namespace'),
```
TheBigLee, line 633: Why has this patch been removed?

component/component/vshn_redis.jsonnet
```diff
@@ -503,188 +503,190 @@ local composition =
...
-            name: 'redis-func',
-            type: 'Container',
-            config: kube.ConfigMap('xfn-config') + {
+            step: 'patch-and-transform',
+            functionRef: {
+              name: 'function-patch-and-transform',
+            },
+            input: {
+              apiVersion: 'pt.fn.crossplane.io/v1beta1',
+              kind: 'Resources',
```
TheBigLee, line 516: Why has the P&T handling been changed here for redis but not for PostgreSQL?

zugao (approved):
LGTM but I have not tested it.

component/class/defaults.yml
Kidswiss: This is now only used for the proxy in debugging, the endpoint is completely crossplane managed.

component/component/provider.jsonnet
Kidswiss: Because Crossplane 1.14 will automatically deploy the SA. It's still referenced in the `rolebinding` because we want our roles to be applied to it.

component/component/vshn_postgres.jsonnet
Kidswiss: Because this patch was always broken, it tries to patch something that doesn't exist. Previous Crossplane versions just discarded a broken patch, the new function based implementation is stricter.

```
cannot compose resources: cannot apply composed resource "sg-backup": failed to create typed patch object (/pgsql-app1-prod-ggtd4-object-storage; kubernetes.crossplane.io/v1alpha1, Kind=Object): .spec.forProvider.spec: field not declared in schema
```

component/component/vshn_redis.jsonnet
TheBigLee: Ah nevermind. Reading a diff is hard ;-)

TheBigLee (approved):

PR #267, opened 2023-11-30
Add e2e tests for our VSHN services

TheBigLee:
## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/class/defaults.yml
+++ b/component/class/defaults.yml
@@ -6,7 +6,7 @@ parameters:
         version: 17.7.1
       mariadb:
         source: https://charts.bitnami.com/bitnami
-        version: 10.1.3
+        version: 11.6.2
       minio:
         source: https://charts.min.io
         version: 5.0.13
--- a/component/component/vshn_postgres.jsonnet
+++ b/component/component/vshn_postgres.jsonnet
@@ -424,6 +424,12 @@ local sgCluster = {
                   },
                   postgres: {
                     version: '',
+                    extensions: [
+                      {
+                        name: 'pg_repack',
+                        version: '1.5.0',
+                      },
+                    ],
                     ssl: {
                       enabled: true,
                       certificateSecretKeySelector: {
--- a/component/component/vshn_postgres.jsonnet
+++ b/component/component/vshn_postgres.jsonnet
@@ -424,12 +424,6 @@ local sgCluster = {
                   },
                   postgres: {
                     version: '',
-                    extensions: [
-                      {
-                        name: 'pg_repack',
-                        version: '1.5.0',
-                      },
-                    ],
                     ssl: {
                       enabled: true,
                       certificateSecretKeySelector: {
--- a/tests/e2e/keycloak/01-connect.yaml
+++ b/tests/e2e/keycloak/01-connect.yaml
@@ -12,10 +12,10 @@ spec:
       restartPolicy: Never
       containers:
         - name: connect
-          image: docker.io/appuio/oc:v4.14
+          image: curlimages/curl:8.6.0
           imagePullPolicy: IfNotPresent
           command:
-            - bash
+            - /bin/sh
           args:
             - -c
             - echo "Testing health...\n" && sleep 30 && curl -ivv http://$KEYCLOAK_HOST/health
--- a/tests/e2e/keycloak/01-connect.yaml
+++ b/tests/e2e/keycloak/01-connect.yaml
@@ -1,0 +1,24 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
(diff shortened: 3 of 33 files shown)


Kidswiss (changes requested):
There seems to be a kuttl test with wrong identations.

tests/e2e/keycloak/01-connect.yaml
```diff
@@ -0,0 +1,24 @@
...
+  backoffLimit: 20
+  template:
+    metadata:
+      labels:
+        e2e-test: keycloak-e2e
+    spec:
+      restartPolicy: Never
+      containers:
+        - name: connect
+          image: docker.io/appuio/oc:v4.14
```
Kidswiss, line 15: This image does not work on Macs with Apple silicon: `no match for platform in manifest: not found`. Not sure why though, as other x86 images usually work with no problem.

However, as you only need curl, maybe some other image would be better suited here? Maybe some small alpine image with curl installed.

Kidswiss (changes requested):
Some things to adjust.

component/class/defaults.yml
```diff
@@ -6,7 +6,7 @@ parameters:
         version: 17.7.1
       mariadb:
         source: https://charts.bitnami.com/bitnami
-        version: 10.1.3
+        version: 11.6.2
```
zugao, line 9: Why do we update the chart version in an e2e PR?

tests/e2e/keycloak/01-connect.yaml
zugao: Agree here. I also use a mac and a lighter image is welcomed.

tests/e2e/keycloak/01-connect.yaml
```diff
@@ -0,0 +1,24 @@
...
+          imagePullPolicy: IfNotPresent
+          command:
+            - bash
+          args:
+            - -c
+            - echo "Testing health...\n" && sleep 30 && curl -ivv http://$KEYCLOAK_HOST/health
```
zugao, line 21: I'm not sure if we really need this curl. I guess the deployment is already checking this same endpoint. As soon as the deployment is ready then it is reflected in the status field object thus making this job obsolete? @Kidswiss what do you think? I would rather do other checks instead of curl if we want to be sure the intance is available from outside the deployment management.

zugao (changes requested):
Some small things, rest is good.

component/class/defaults.yml
wejdross: because e2e tests found an issue in old charts :D

component/component/vshn_postgres.jsonnet
```diff
@@ -424,6 +424,12 @@ local sgCluster = {
                   },
                   postgres: {
                     version: '',
+                    extensions: [
```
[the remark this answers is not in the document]
wejdross: deleted that section

tests/e2e/kuttl-test.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestSuite
+testDirs:
+  - ./test/e2e/
+namespace: widera-testing
```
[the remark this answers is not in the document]
wejdross: updated that section

tests/e2e/keycloak/01-connect.yaml
wejdross: switched to another image also mentioned issue with that image on our #appuio-cloud RC channel

tests/e2e/keycloak/01-connect.yaml
wejdross: that was advised by @Kidswiss

wejdross:
@Kidswiss @zugao I just added new directory and file .githooks/pre-commit - it's small and simple python script to prevent user from commiting code when keycloak docker registry credentials are present in code, it's copied into .git/hooks directory whenever someone start compilation of golden files. Unfortunately this is the best protection we can apply because git doesn't like setting such script automatically

tests/e2e/keycloak/01-connect.yaml
Kidswiss: @zugao There's currently no assert for the deployment status, so it's not really obsolete.

Also in the spirit of end-to-end test, this curl does more than the liveness probe of the pod:
* It uses the `KEYCLOAK_HOST`
* It connects from outside -> network policies

We can make it do some authenticated call in a future iteration as well.

Kidswiss (approved):
LGTM

wejdross (approved):

zugao (commented):
LGTM

PR #303, opened 2024-01-16
Adjust PSQL sidecar limit

wejdross:
## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/class/defaults.yml
+++ b/component/class/defaults.yml
@@ -42,7 +42,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.49.0
+        tag: upgrade/stackgres_version
       apiserver:
         registry: ghcr.io
         repository: vshn/appcat-apiserver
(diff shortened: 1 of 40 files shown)


component/class/defaults.yml
```diff
@@ -42,7 +42,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.49.0
+        tag: upgrade/stackgres_version
```
TheBigLee, line 45: This needs to be fixed before merging

TheBigLee (approved):
Please see comment, otherwise LGTM

component/class/defaults.yml
```diff
@@ -457,7 +457,7 @@ parameters:
                 memory: "30Mi"
               limits:
                 cpu: "32m"
-                memory: "2Gi"
+                memory: "64Mi"
```
TheBigLee, line 460: This actually might be too low. Have you verified that it works with this?

Kidswiss: I agree, I would go with 256 or something.

PR #313, opened 2024-01-25
adding non-sla minio alerts

wejdross:
## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/component/vshn_minio.jsonnet
+++ b/component/component/vshn_minio.jsonnet
@@ -36,7 +36,8 @@ local prometheusRule = prom.GeneratePrometheusNonSLORules(
         {
           alert: 'minioDiskOfflineCritical',
           annotations: {
-            description: "Minio server reports it's disk is down - {{ $labels.name }} in namespace {{ $labels.label_appcat_vshn_io_claim_namespace }}.",
+            description: "Minio server reports it's disk is down - {{ $labels.name }} in namespace {{ $labels.label_appcat_vshn_io_claim_namespace }}.
+            ",
             summary: 'Minio disk is offline for at least 30 seconds.',
           },
 
@@ -106,6 +107,10 @@ local composition =
             },
           },
           {
+            step: 'patch-and-transform',
+            functionRef: {
+              name: 'function-patch-and-transform',
+            },
             input: {
               apiVersion: 'pt.fn.crossplane.io/v1beta1',
               kind: 'Resources',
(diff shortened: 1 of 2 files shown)


component/component/vshn_minio.jsonnet
```diff
@@ -25,6 +25,39 @@ local connectionSecretKeys = [
...
+      name: 'minio-diskOffline',
+      rules: [
+        {
+          alert: 'minioDiskOfflineCritical',
+          annotations: {
+            description: "Minio server reports it's disk is down - {{ $labels.name }} in namespace {{ $labels.label_appcat_vshn_io_claim_namespace }}.",
+            summary: 'Minio disk is offline for at least 30 seconds.',
+          },
+
+          expr: 'minio_cluster_drive_offline_total != 0',
```
Kidswiss, line 43: Do we actually scrape this? Minio's metrics endpoint is behind authentication.

But it would be a very good alert nonetheless. So I would keep it either way. If/when we add the Minio metrics in the future will already have the alert.

Kidswiss (approved):
Just one question to clarify, the rest LGTM

component/component/vshn_minio.jsonnet
wejdross: actually we're scraping those metrics thanks to @TheBigLee 
https://github.com/vshn/appcat/blob/master/pkg/comp-functions/functions/vshnminio/minio_deploy.go#L271

how it works without auth - no idea :D 

we're able to access those metrics (at least most of them) via:
https://console.lab-cloudscale-rma-0.appuio.cloud/monitoring/query-browser?query0=minio_cluster_drive_offline_total

Minio inside metrics gives us great potential for Grafana charts as it provides a lot of valuable insides, but very little for active alerting imho

Kidswiss: Even better! I thought we didn't have this yet because of the auth, but I like to be surprised positively :D

Kidswiss:
I just realized that Minio was actually a comp function only composition so far.

So it makes more sense to also deploy the non-slo-alerts via the comp functions.

Kidswiss:
you can probably close this in favor of https://github.com/vshn/appcat/pull/128

TheBigLee (commented):
Why are we doing those rules with PnT and others with CompFunctions?
This looks very inconsistent to me

PR #317, opened 2024-02-07
adjusting psql sidecar memory

wejdross:
## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/class/defaults.yml
+++ b/component/class/defaults.yml
@@ -467,7 +467,7 @@ parameters:
                 memory: "30Mi"
               limits:
                 cpu: "32m"
-                memory: "2Gi"
+                memory: "256Mi"
             runDbops:
               requests:
                 cpu: "100m"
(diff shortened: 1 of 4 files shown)


Kidswiss (commented):
Have you checked on the lab that these values don't violate the ratio?

wejdross:
how can I check that?

Kidswiss:
> how can I check that?

```
k -n instanceNamespace describe pod postgresqlpod
```

Then there will be an event that looks something like this:
```
Events:
  Type     Reason             Age                    From                       Message
  ----     ------             ----                   ----                       -------
  Warning  TooMuchCPURequest  2m59s (x2935 over 9h)  resource-ratio-controller  Current memory to CPU ratio of 3692Mi/core for node type "appuio.io/node-class=plus,node-role.kubernetes.io/app=" in this namespace is below the fair use ratio of 4Gi/core. APPUiO Cloud bills CPU requests which exceed the fair use ratio. See https://vs.hn/appuio-cloud-cpu-requests for instructions to adjust the requests.
```

component/class/defaults.yml
```diff
@@ -467,7 +467,7 @@ parameters:
                 memory: "30Mi"
               limits:
                 cpu: "32m"
-                memory: "2Gi"
+                memory: "256Mi"
```
TheBigLee, line 470: I assume 128Mi is not enough?

TheBigLee (commented):
And as Simon already said, please ensure that the ratio is correct

TheBigLee (approved):

PR #334, opened 2024-03-26
Fix jq version for github actions

Kidswiss:
## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/component/app.jsonnet
+++ b/component/component/app.jsonnet
@@ -3,7 +3,23 @@ local inv = kap.inventory();
 local params = inv.parameters.appcat;
 local argocd = import 'lib/argocd.libjsonnet';
 
-local app = argocd.App('appcat', '');
+local app = argocd.App('appcat', '') + (
+  if params.services.vshn.e2eTests then {
+    spec+: {
+      ignoreDifferences+: [
+        {
+          group: '',
+          jsonPointers: [
+            '/data',
+          ],
+          kind: 'Secret',
+          name: 'github-ci-secret',
+          namespace: 'appcat-e2e',
+        },
+      ],
+    },
+  } else {}
+);
 
 {
   appcat: app,
--- a/component/component/rbac_e2e.jsonnet
+++ b/component/component/rbac_e2e.jsonnet
@@ -26,6 +26,21 @@ local e2eRoleBinding = kube.RoleBinding('appcat-e2e') + {
   subjects_: [ e2eSA ],
 };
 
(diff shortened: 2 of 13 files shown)


component/component/rbac_e2e.jsonnet
```diff
@@ -4,39 +4,54 @@ local kube = import 'lib/kube.libjsonnet';
...
 };
 
-local e2eRoleBinding = kube.RoleBinding('appcat-e2e') + {
+local e2eRoleBinding = kube.RoleBinding('schedar-e2e') + {
   metadata+: {
-    namespace: 'appcat-e2e',
+    namespace: 'schedar-e2e',
   },
-  roleRef_: kube.ClusterRole('admin'),
+  roleRef_: kube.ClusterRole('gitlab-ci'),
```
wejdross, line 25: github-ci

Kidswiss: I just realized that they both don't exist... I guess I just use `admin` here. It's also what the docs APPUiO use.

component/component/app.jsonnet
```diff
@@ -3,7 +3,23 @@ local inv = kap.inventory();
...
+    spec+: {
+      ignoreDifferences+: [
+        {
+          group: '',
+          jsonPointers: [
+            '/data',
+          ],
+          kind: 'Secret',
+          name: 'github-ci-secret',
+          namespace: 'schedar-e2e',
```
zugao, line 17: I would keep the project name instead of team name.

wejdross: I wouldn't - it's descriptive for all engineers right now, it's also easy way to contact with us.

Kidswiss: It's currently deployed as `schedar-e2e`, this adjustment makes that the jsonnet actually matches what's deployed.

But if you like the old name more, I can change it.

wejdross (approved):
looks good, I saw it worked for LAB cluster, I don't care that much about namespace name, LGTM

component/component/app.jsonnet
Kidswiss: > I wouldn't - it's descriptive for all engineers right now, it's also easy way to contact with us. Component can be owned by different teams, `schedar` is just one

True, but schedar owns appcat, so it should be the same either way.

But thinking a bit more about the naming, it might make sense to go with `appcat-e2e`. Otherwise, we're leaking some vshn internal naming to the product.

PR #336, opened 2024-04-03
Update e2e.yaml

wejdross:
it might be possible that everyone run some jobs at 00:00 and our e2e test times out because of that. I suggest change of time to see if that's the issue


wejdross:
closing after talk with @Kidswiss we should first update provider-kubernetes

PR #344, opened 2024-04-18
Add/keycloak metrics

wejdross:
- Keycloak metrics are now enabled and scraped
- Keycloak 24 is available alongside with Keycloak 23
- Keycloak 24 has image with automatically configured and enabled metrics

### DIFF
--- a/component/class/defaults.yml
+++ b/component/class/defaults.yml
@@ -45,7 +45,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.63.1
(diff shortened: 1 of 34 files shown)


component/class/defaults.yml
```diff
@@ -45,7 +45,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.63.1
+        tag: add/keycloak_metrics
```
wejdross, line 48: all of those entries will be adjusted

Kidswiss (approved):

PR #353, opened 2024-05-02
Manage netpol in postgres

zugao:
This PR help to enable custom namespace access to all vshn services via netpols in comp-functions. We do not clean the existing netpols for redis, minio and postgresql in this PR in order to avoid access permission problem to the service when deploying this PR. I will create a new PR to clean the old netpols when this one is already deployed.

## Checklist

      changelog.
      linked in the changelog.


component/class/defaults.yml
```diff
@@ -45,7 +45,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.66.2
+        tag: allow-namespaces
```
zugao, line 48: This will be changed

Kidswiss (approved):

PR #371, opened 2024-05-30
Add usermanagement for PostgreSQL

Kidswiss:
We have to enable Crossplane's usages in order for this feature to work correctly.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/class/defaults.yml
+++ b/component/class/defaults.yml
@@ -34,6 +34,10 @@ parameters:
         registry: ghcr.io
         repository: vshn/provider-minio/provider
         tag: v0.1.0
+      provider-sql:
+        registry: xpkg.upbound.io
+        repository: crossplane-contrib/provider-sql
+        tag: v0.9.0
       sloth:
         registry: ghcr.io
         image: slok/sloth
@@ -45,7 +49,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.72.0
+        tag: v4.73.0
       apiserver:
         registry: ghcr.io
         repository: vshn/appcat-apiserver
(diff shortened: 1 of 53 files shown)


component/class/defaults.yml
```diff
@@ -45,7 +49,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.71.2
+        tag: add/user_management
```
Kidswiss, line 52: Will be adjusted once there's a tag

wejdross (approved):

component/class/defaults.yml
```diff
@@ -402,6 +406,16 @@ parameters:
         providerConfigRefs: []
         additionalProviderConfigs: []
 
+      sql:
+        enabled: false
```
zugao, line 410: maybe we should not exposed this parameter? `provider-sql` must be available for services that expose user management. What would happen if the sql provider is disabled meanwhile the service postgres allows user management?

component/component/provider.jsonnet
```diff
@@ -383,4 +393,47 @@ local runtimeConfigRef(name) = {
       role,
       rolebinding,
     ],
+  [if params.providers.sql.enabled || params.services.vshn.postgres.enabled then '10_provider_sql']:
```
zugao, line 396: I would not use `params.providers.sql.enabled` all together. Just enabled provider sql for any service that needs it.

zugao (commented):
Some things to consider

component/component/provider.jsonnet
Kidswiss: Fine by me

zugao (approved):
LGTM

PR #406, opened 2024-07-10
Fix compile error if no openshift template is specified

TheBigLee:
## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/tests/vshn.yml
+++ b/component/tests/vshn.yml
@@ -152,7 +152,7 @@ parameters:
                   appuio.io/node-class: "plus"
               note: "Will be scheduled on APPUiO Cloud plus nodes"
         minio:
-          enabled: false
+          enabled: true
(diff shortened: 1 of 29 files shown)


component/tests/vshn.yml
```diff
@@ -152,7 +152,7 @@ parameters:
                   appuio.io/node-class: "plus"
               note: "Will be scheduled on APPUiO Cloud plus nodes"
         minio:
-          enabled: false
+          enabled: true
```
Kidswiss, line 155: Can you reset this to false? For local dev a minio instance will already be provisioned via kindev, so we can save a bit of resource by not running minio twice.

Kidswiss (changes requested):

zugao (approved):
LGTM

PR #407, opened 2024-07-11
Add very rudimentary backup monitoring

Kidswiss:
This adds a backup monitoring via jobs. It alerts if there's a job that has failed in any of the AppCat namespaces.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/class/defaults.yml
+++ b/component/class/defaults.yml
@@ -438,8 +438,6 @@ parameters:
           smtpFromAddress: myuser@example.com
           secretNamespace: syn-appcat
           secretName: mailgun-smtp-credentials
-        stsResizer:
-          enabled: true
         postgres:
           billing: true
           # bucket_region: 'lpg' || 'ch-gva-2'
--- a/component/component/main.jsonnet
+++ b/component/component/main.jsonnet
@@ -195,12 +195,48 @@ local emailSecret = kube.Secret(params.services.vshn.emailAlerting.secretName) {
   },
 };
 
+local filterName(name) = '|' + if name == 'postgres' then 'postgresql' else if name == 'emailAlerting' then '' else name;
+local backupJobRegex = std.foldl(function(prev, current) prev + filterName(current.name), common.FilterServiceByBoolean('enabled'), '');
+
+local backupPrometheusRule = {
+  apiVersion: 'monitoring.coreos.com/v1',
+  kind: 'PrometheusRule',
+  metadata: {
+    name: 'appcat-backup',
+    namespace: params.namespace,
+  },
+  spec: {
+    groups: [
+      {
+        name: 'appcat-backup',
+        rules: [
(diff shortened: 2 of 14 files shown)


zugao (commented):
Why don't we add this in the appcat common package?

Kidswiss:
> Why don't we add this in the appcat common package?

Common is a library and doesn't produce any files by itself.

component/tests/golden/apiserver/appcat/appcat/10_appcat_backup_monitoring.yaml
```diff
@@ -0,0 +1,21 @@
...
+  groups:
+    - name: appcat-backup
+      rules:
+        - alert: AppCatBackupJobError
+          annotations:
+            description: The backup job {{ $labels.job_name }} in namespace {{ $labels.namespace
+              }} has failed.
+            runbook_url: https://kb.vshn.ch/app-catalog/how-tos/appcat/AppCatBackupJobError.html
+            summary: AppCat service backup failed.
+          expr: kube_job_failed{job_name=~".*backup.*", namespace=~"vshn-(postgres|redis|stsResizer)-.*"}
```
zugao, line 16: Shouldn't it be postgresql instead of posgres? Why there's stsResizer as well?

Kidswiss: As discussed in chat, I will refactor the `services.vshn` object, so it only contains actual vshn services.

component/component/main.jsonnet
```diff
@@ -195,12 +195,48 @@ local emailSecret = kube.Secret(params.services.vshn.emailAlerting.secretName) {
   },
 };
 
+local filterName(name) = '|' + if name == 'postgres' then 'postgresql' else if name == 'emailAlerting' then '' else name;
+local backupJobRegex = std.foldl(function(prev, current) prev + filterName(current.name), common.FilterServiceByBoolean('enabled'), '');
```
TheBigLee, line 199: This regex currently produces strange (but probably valid) output in some cases. eg:
```
vshn-(||minio)-
```

component/class/defaults.yml
```diff
@@ -438,8 +438,6 @@ parameters:
           smtpFromAddress: myuser@example.com
           secretNamespace: syn-appcat
           secretName: mailgun-smtp-credentials
-        stsResizer:
-          enabled: true
```
wejdross, line 442: why is this removed here?

component/component/main.jsonnet
wejdross: it allows empty value:
```
vshn--
```
is a match. that's bad. I know it's a corner case we wouldn't probably ever get, but I'd fix it just to sleep better

PR #420, opened 2024-07-19
Various fixes for the e2e tests

Kidswiss:
* Wrong asserts for Nextcloud
* Wrong connection job for Nextcloud
* Fix race condition concerning the ingress check. It got moved to a
  later assert, is it might take a while to provision
* Reduced Parallelism to reduce resource starvation on the lab

They are running through now: https://github.com/vshn/component-appcat/actions/runs/10037795049/job/27738357045

## Checklist

      changelog.
      linked in the changelog.


tests/e2e/nextcloud/00-assert.yaml
```diff
@@ -20,13 +20,9 @@ spec:
         instances: 1
       relativePath: /
       serviceLevel: besteffort
-      version: "23"
-      #version: 23.0.5-202402021353-44-af6cea11
+      version: "29"
```
zugao, line 23: Do we even need to test the version? Will have to update it all the time. If we can omit the version from the claim we  can remove version from testing.

zugao (approved):
LGTM but added a note

tests/e2e/nextcloud/00-assert.yaml
Kidswiss: Fine by me

wejdross (approved):

PR #429, opened 2024-07-30
Add support for Keycloak 25

TheBigLee:
This PR adds support for Keycloak 25 and drops support for older versions of Keycloak due to breaking changes in Keycloak and the helm chart.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/class/defaults.yml
+++ b/component/class/defaults.yml
@@ -12,7 +12,7 @@ parameters:
         version: 5.0.13
       keycloak:
         source: https://codecentric.github.io/helm-charts
-        version: 2.3.0
+        version: 2.4.3
       nextcloud:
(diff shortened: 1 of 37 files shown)


component/class/defaults.yml
```diff
@@ -52,7 +52,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.86.0
+        tag: support_keycloak_25
```
Kidswiss, line 55: Don't forget to adjust once it's released

Kidswiss (approved):

component/class/defaults.yml
TheBigLee: Done

PR #494, opened 2024-09-27
Add providerConfig management

vshnbot:
## Summary

* Each claim/composite pair can specific their own providerConfig name
* It will be injected into each managed resource
* The providerConfigs need to be correctly provisioned beforehand
* Not setting the label will result in the current defaults being used
* Some slight refactoring for the Minio kindev configuration

## Checklist

- [x] Update tests.

Link: https://github.com/vshn/appcat/pull/237


tests/golden/apiserver/appcat/appcat/10_provider_sql.yaml
```diff
@@ -45,18 +45,6 @@ metadata:
     name: crossplane-provider-provider-sql-system-custom
   name: crossplane:provider:provider-sql:system:custom
 rules:
-  - apiGroups:
```
Kidswiss, line 48: This rule is invalid and never actually worked. Thus it's not necessary.

docs/modules/ROOT/pages/references/component-parameters.adoc
```diff
@@ -67,34 +67,106 @@ reference:: xref:references/slo-parameters.adoc[See SLO parameters]
...
+required:: true
+
+The namespace of the secret.
+
+==== `data`
+type:: map
+default:: `{}`
+required:: true
+
+The contents of the `data` map of the secret.
```
zugao, line 127: I guess here you will have the details on how to connect to a different cluster? If yes can you elaborate more?

zugao (approved):
LGTM

docs/modules/ROOT/pages/references/component-parameters.adoc
Kidswiss: This varies from provider to provider. So you'd have to consult the provider's docs.

PR #511, opened 2024-10-18
Add Collabora integration for Nextcloud

wejdross:
Automatic installation of Collabora with splitted logic for Openshift and Kind clusters
Collabora works with SSL, Ingress/Route calls it using SSL, customers connects to Ingress/Route using SSL


Makefile
```diff
@@ -94,4 +94,26 @@ clean: ## Clean the project
 
 .PHONY: pre-commit-hook
 pre-commit-hook: ## Install pre-commit hook in .git/hooks
-	/usr/bin/cp -fa .githooks/pre-commit .git/hooks/pre-commit
\ No newline at end of file
+	/usr/bin/cp -fa .githooks/pre-commit .git/hooks/pre-commit
+
+.PHONY: local-appcat-development
```
zugao, line 99: Can you add some comments here what these targets do? It's not clear. Also this has nothing to do with this PR but it's ok for me as long as this is documented what it does and why it exists otherwise nobody will know about it let alone use it.

component/provider.jsonnet
```diff
@@ -164,10 +164,21 @@ local providerRBAC = {
...
+      {
+        apiGroups: [ '' ],
+        resources: [ 'persistentvolumeclaims' ],
         verbs: [ 'get', 'list', 'watch', 'create', 'watch', 'patch', 'update', 'delete' ],
       },
+      {
+        // This is needed for the privileged SCC - Nextcloud -> Collabora
+        apiGroups: [ 'security.openshift.io' ],
+        resources: [ 'securitycontextconstraints' ],
+        verbs: [ '*' ],
```
zugao, line 180: This is a bit tricky here and will downgrade our security significantly. Do we really need `*` here?

component/vshn_appcat_services.jsonnet
```diff
@@ -179,6 +179,42 @@ local vshn_appcat_service(name, serviceParams) =
...
+      allowHostIPC: true,
+      allowHostNetwork: true,
+      allowHostPID: true,
+      allowHostPorts: true,
+      allowPrivilegeEscalation: false,
+      allowPrivilegedContainer: true,
+      allowedCapabilities: [
+        'MKNOD',
+        'CHOWN',
+        'SYS_CHROOT',
```
zugao, line 195: Can you please explain why do we need root capability? We need to be very careful with this resource and make sure we give the least amount of privelegies.

zugao (commented):
Added some comments

component/vshn_appcat_services.jsonnet
wejdross: it's CHROOT - that's different thing https://man7.org/linux/man-pages/man2/chroot.2.html

zugao: Can you please write here where these capabilities are required? Or we can have a short call.

Makefile
wejdross: I removed that logic as even I stopped using that

component/provider.jsonnet
```diff
@@ -54,8 +54,8 @@ local providerRBAC = {
       },
       {
         apiGroups: [ '' ],
-        resources: [ 'namespaces', 'serviceaccounts', 'secrets', 'pods', 'pods/log', 'pods/portforward', 'pods/status', 'services' ],
-        verbs: [ 'get', 'list', 'watch', 'create', 'watch', 'patch', 'update', 'delete' ],
+        resources: [ 'namespaces', 'serviceaccounts', 'secrets', 'pods', 'pods/log', 'pods/portforward', 'pods/status', 'pods/attach', 'pods/exec', 'services' ],
+        verbs: [ '*' ],
```
Kidswiss, line 58: Please don't use `*` for the verbs.

class/defaults.yml
```diff
@@ -867,6 +867,7 @@ parameters:
           restoreSA: nextcloudserviceaccount
           restoreRoleRules: ${appcat:defaultRestoreRoleRules}
           additionalInputs:
+            collabora_image: "collabora/code:24.04.8.1.1"
```
Kidswiss, line 870: Please don't hardcode the image here, do it the same way we do it for all other images.

Kidswiss (changes requested):

component/vshn_appcat_services.jsonnet
```diff
@@ -187,6 +223,7 @@ local vshn_appcat_service(name, serviceParams) =
     ['20_plans_vshn_%s' % name]: plansCM,
     ['22_prom_rule_sla_%s' % name]: promRuleSLA,
     [if isOpenshift && std.objectHas(serviceParams, 'openshiftTemplate') then '21_openshift_template_%s_vshn' % name]: osTemplate,
+    [if isOpenshift then '22_scc_appcat']: scc,
```
Kidswiss, line 226: Can you move this out of the generic `vshn-services` logic? Right now it will always deploy this specific collabora manifest if it's openshift, even if collabora is not enabled on a cluster. I'd probably put it in a `vshn_nextcloud.jsonnet` file.

wejdross: I just added additional variable to if statement to verify whether nextcloud is enabled or not. separate file to just handle scc seems like an overengineering

Makefile
```diff
@@ -94,4 +94,4 @@ clean: ## Clean the project
 
 .PHONY: pre-commit-hook
 pre-commit-hook: ## Install pre-commit hook in .git/hooks
-	/usr/bin/cp -fa .githooks/pre-commit .git/hooks/pre-commit
```
zugao, line 97: Would be nice to have the `trailing newline` here.

class/defaults.yml
```diff
@@ -866,10 +870,13 @@ parameters:
           restoreSA: nextcloudserviceaccount
           restoreRoleRules: ${appcat:defaultRestoreRoleRules}
           additionalInputs:
+            collabora_image: ${appcat:images:collabora:registry}/${appcat:images:collabora:image}:${appcat:images:collabora:tag}
+            collaboraCPULimit: "1"
+            collaboraCPURequests: 250m
+            collaboraMemoryLimit: 1Gi # during my tests I was able to force collabora to use ~800Mi
```
zugao, line 876: The comment does not help much here.

zugao (approved):
LGTM but added some nitpicks

component/vshn_appcat_services.jsonnet
wejdross: I just switched logic to separate vshnnextcloud.jsonnet file.

zugao: I agree with @Kidswiss, the `scc` file should go to `vshn-nextcloud.jsonnet`. It's unfortunate that we don't have it for nextcloud yet but we have a `.jsonnet` for every other service. It's time to give nextcloud it's own file.

wejdross: It's already in separate file.

Kidswiss (approved):

PR #512, opened 2024-10-21
PR for appcat on change-billing

vshnbot:
## Summary

Currently we are getting data for billing from the workloads (deployments and statefulsets) special label `appcat.io/billing=true`. This approach proved to be ineffective for some services such as VSHNPostgreSQL which is managed by Stackgres. Unfortunately the operator sometimes is managing the pods directly, by passing the workload. In this particular case we end up without the necessary label in the pod definition.

This new approach manages differently the billing architecture. If previously we were having one prometheus rule which would gather unnecessarily difficult specific data, now we have one rule per each instance namespace and it is managed directly by the composition function where we get reliable data from the source.

## Checklist

- [x] Update tests.

Link: https://github.com/vshn/appcat/pull/246


class/defaults.yml
```diff
@@ -55,7 +55,7 @@ parameters:
       appcat:
         registry: ghcr.io
         repository: vshn/appcat
-        tag: v4.98.0
+        tag: change-billing
```
Kidswiss, line 58: Please don't break it with `git push --force` :D

Kidswiss (approved):
LGTM

But please be sure that the metrics on mimir are identical. Ideally roll it out to the lab and run the odoo job then verify with tobru.

TheBigLee (approved):

tests/vshn.yml
```diff
@@ -22,7 +22,7 @@ parameters:
   appcat:
 
     grpcEndpoint: host.docker.internal:9443
-    proxyFunction: false
+    proxyFunction: true
```
Kidswiss, line 25: Please revert

component/billing.jsonnet
```diff
@@ -157,8 +157,8 @@ local generateCloudAndManaged = function(name)
   // For postgresql we have a missmatch between the label and the name in our definition.
   local queryName = if name == 'postgres' then name + 'ql' else name;
 
-  local managedQuery = 'appcat:metering{label_appuio_io_billing_name="appcat-' + queryName + '",label_appcat_vshn_io_sla="%s", tenant_name!="APPUiO"}';
-  local cloudQuery = 'appcat:metering{label_appuio_io_billing_name="appcat-' + queryName + '",label_appcat_vshn_io_sla="%s", tenant_name="APPUiO"} * on(label_appuio_io_organization) group_left(sales_order) label_replace(appuio_control_organization_info{namespace="appuio-control-api-production"}, "label_appuio_io_organization", "$1", "organization", "(.*)")';
+  local managedQuery = 'sum_over_time(appcat:metering{label_appuio_io_billing_name="appcat-' + queryName + '",label_appcat_vshn_io_sla="%s", tenant_name!="APPUiO"}[60:1])/60';
```
Kidswiss, line 160: I just remembered something:
When we do this exact query via the golang prometheus client, it will be slightly off compared to the GUI and recording rules.

For some weird obscure reason we have to use `[59:1]` instead of ´[60:1]´.

Refs:
Aldebaran has that in every of their cronjobs: https://github.com/vshn/package-appuio-reporting-aldebaran/blob/master/main.yml#L1015
Note about this the legacy AppCat query: https://github.com/vshn/component-appcat/blob/master/component/promql/appcat.promql#L63

Kidswiss (changes requested):
One more thing and then LGTM

component/billing.jsonnet
zugao: Interesting

Kidswiss (approved):

PR #534, opened 2024-11-13
Enable mail configuration by default

Kidswiss:
The settings are adjusted to match APPUiO Cloud. According to internal docs our mailgun currently only supports the `appuio.cloud` domain.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/class/defaults.yml
+++ b/class/defaults.yml
@@ -501,11 +501,11 @@ parameters:
       controlNamespace: "syn-appcat-control"
       # Works only for VSHN services for now
       emailAlerting:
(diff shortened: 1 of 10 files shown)


zugao (approved):
LGTM. The email is disabled as long as customer does not add his email, is that right?

Kidswiss:
> LGTM. The email is disabled as long as customer does not add his email, is that right?

Yes that is correct.

Also, we might want to have another sender address in the future, but for now this is good enough... At least they should get mails with this.

PR #536, opened 2024-11-15
Add Control Plane support

zugao:
This PR helps to separate manifests depending cluster management system. At this stage it cannot separate precisely each manifest as this will be decided further when we adjust various applications.

## Checklist

      changelog.
      linked in the changelog.


postprocess/crossplane.jsonnet
```diff
@@ -0,0 +1,26 @@
...
+ * Remove generated crossplane manifests if not required
+ */
+local com = import 'lib/commodore.libjsonnet';
+
+local inv = com.inventory();
+local params = inv.parameters.appcat;
+local cms = params.clusterManagementSystem;
+local file_extension = '.yaml';
+
+local isSingleCluster = cms.controlPlaneCluster && cms.serviceCluster;
```
Kidswiss, line 11: can't you import the vars file here as well?

postprocess/crossplane.jsonnet
```diff
@@ -0,0 +1,26 @@
+/**
+ * Remove generated crossplane manifests if not required
+ */
+local com = import 'lib/commodore.libjsonnet';
+
+local inv = com.inventory();
+local params = inv.parameters.appcat;
+local cms = params.clusterManagementSystem;
```
Kidswiss, line 8: Can't you do this here?

```suggestion
local cms = com.inventory().parameters.appcat.clusterManagementSystem;
```

Kidswiss (approved):
LGTM

However, I wonder if we can streamline this a bit more in the future? It's just a matter of time someone forgets to set the right if clause for some new functionality...

zugao:
> However, I wonder if we can streamline this a bit more in the future? It's just a matter of time someone forgets to set the right if clause for some new functionality...

I agree but at the moment I have no idea how we could streamline this better. In the end the engineer will still have to decide himself where each manifests should end up.

postprocess/crossplane.jsonnet
zugao: I had a problem here with importing the `vars.jsonnet`

postprocess/crossplane.jsonnet
zugao: I will move this change in the next PR when I resolve the importing issues.

postprocess/crossplane.jsonnet
zugao: I will move this change in the next PR when I resolve the importing issues.

PR #544, opened 2024-11-19
Consolidate golden tests

zugao:
Consolidated the golden tests. The running time of target `gen-golden-all` went down from 5m10s to 3m10s.

* This PR removes unnecessary golden tests and adds the configuration in both `vshn-cloud` and `vshn-managed` golden tests.
* The Cluster Management System also has it's own 2 tests. 
* We keep the `defaults` and use the old `vshn` golden test as `dev` with `proxyFunction` set to `true`.
* Removed make `test` target from the Pipeline as it was doing the same as `diff`

## Checklist

      changelog.
      linked in the changelog.


Kidswiss (approved):
Yeah, just leaving the vshn/dev golden as is 👍

PR #555, opened 2024-11-26
Fix missing restAPI when using stackgres with OLM

TheBigLee:
We need the Stackgres RestAPI for our maintenance. Without the RestAPI we won't be able to find the latest minor versions available.

Unfortunately the RestAPI is not installed by default when using OLM and needs to be manually patched in the SGConfig.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/vshn_postgres.jsonnet
+++ b/component/vshn_postgres.jsonnet
@@ -40,6 +40,84 @@ local stackgresOperatorNs = kube.Namespace(params.stackgres.namespace) {
   },
 };
 
+local sa = kube.ServiceAccount('stackgres-sgconfig-restapi-patcher') {
+  metadata+: {
+    namespace: params.stackgres.namespace,
+  },
+};
+
+local role = kube.Role('stackgres-sgconfig-restapi-patcher') {
+  metadata+: {
+    namespace: params.stackgres.namespace,
+  },
+  rules: [
+    {
+      apiGroups: [
+        'stackgres.io',
+      ],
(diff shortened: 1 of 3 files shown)


Kidswiss:
Did you test if our minorVersionUpgrades will work with that?

If I remember correctly the credentials for the REST API are expected to be in a specific secret. If that secret now also looks differently, then the maintenance will still not work.

TheBigLee:
> Did you test if our minorVersionUpgrades will work with that?
> 
> If I remember correctly the credentials for the REST API are expected to be in a specific secret. If that secret now also looks differently, then the maintenance will still not work.

Yes I tested it and the minor upgrade got triggered:
```
NAME                                     READY   STATUS      RESTARTS   AGE
databasesrepack-rvw69                    0/2     Completed   0          35h
minorupgrade-f428s                       0/1     Error       0          35h
minorupgrade-h76w9                       0/1     Error       0          35h
scale-test-qz26z-0                       5/5     Running     0          35h
scale-test-qz26z-backup-28877715-tdgqj   0/1     Completed   0          30h
scale-test-qz26z-backup-28879155-gxckt   0/1     Completed   0          6h7m


However, I just saw that the upgrade fails (this is a different issue though):

TheBigLee:
> > However, I just saw that the upgrade fails (this is a different issue though):
> 
> Did you forget to paste the logs? :D

No. I just did a colon by mistake :-)

PR #556, opened 2024-11-27
Add support for remote control plane to SLI exporter

vshnbot:
## Summary

This adds support for remote control planes for the SLI exporter. It now
takes an additional `KUBECONFIG` env var to connect to the control plane.

Additionally to determine the state of the maintenance of the service
cluster, it connects to the local service cluster to reconcile on
`UpgradeJob` objects.

Also the SLI-Prober doesn't need any configuration about the services
anymore. It will detect these on startup.

## Checklist

- [x] Update tests.

Link: https://github.com/vshn/appcat/pull/273


class/defaults.yml
```diff
@@ -105,6 +105,8 @@ parameters:
     clusterManagementSystem:
       controlPlaneCluster: true
       serviceCluster: true
+      # This value should only be non-empty if control and service plane split is used
```
zugao, line 108: In my opinion this comment should not exist, this gets things more complicated. We need to use variables from `config/vars.jsonnet` to figure out if the component is in split mode. If we add a comment to every new variable here, it's gonna become a mess. I am also not very happy to use those variable everywhere but an we find a solution?

Makefile
```diff
@@ -98,12 +98,22 @@ pre-commit-hook: ## Install pre-commit hook in .git/hooks
...
-	git remote add origin http://gitea_admin:admin@forgejo.127.0.0.1.nip.io:8088/gitea_admin/appcat.git && \
-	git push -u origin master --force
+	git remote add origin http://gitea_admin:adminadmin@forgejo.127.0.0.1.nip.io:8088/gitea_admin/$(repo).git && \
+	git push -u origin master --force && \
+	rm -rf .git
+	yq eval-all '. as $$item ireduce ({}; . * $$item )' hack/base_app.yaml tests/golden/$(instance)/appcat/apps/appcat.yaml \
+	| yq '.metadata.name = "$(repo)"' | yq '.spec.source.repoURL = "http://forgejo-http.forgejo.svc:3000/gitea_admin/$(repo)"' \
+	| yq '.spec.destination.server = "$(cluster)"' | kubectl apply -f -
+
+.PHONY: push-non-converged
```
zugao, line 116: what is this?

Kidswiss: This pushes the configuration for a split setup to argocd.

If you have any better ideas how to name it, please feel free to share.

zugao: it's good, could you add this exact comment there?

Kidswiss: I've added it to the readme. I'll also add a help text.

class/defaults.yml
Kidswiss: I've added asserts now, so it will scream at you if you're doing it wrong.

zugao (approved):

PR #602, opened 2025-01-16
Fix slos namespace

zugao:
Thus PR fixes the name of our appcat slos namespace

## Checklist

      changelog.
      linked in the changelog.


wejdross (approved):
Would there be any issues if we migrate appcat-sla bucket? I think we only store data there, but does anything read from it?

zugao:
> Would there be any issues if we migrate appcat-sla bucket? I think we only store data there, but does anything read from it?

I don't think here is the right place. We can discuss this in an already opened ticket or a new one.

PR #651, opened 2025-03-03
Update crossplane dependencies in appcat 

vshnbot:
## Summary

* Short summary of what's included in the PR
* Give special note to breaking changes

## Checklist

- [ ] Update tests.

Link: https://github.com/vshn/appcat/pull/319


wejdross:
@zugao here You need to resolve conflicts first

zugao:
Well, when review are coming slow .. this happens

wejdross (approved):
It doesn't look bad, few things:

- do not forget to update appcat tag
- do not update to make a proper description of the issue

other than that it's OK

Kidswiss:
I just tested something independently and found that the `.status.observedGeneration: field not declared in schema` error also happens for nested postgresql.

Going through the diff it doesn't look like that field gets injected there. Please have a look.

Kidswiss (changes requested):
Okay after some digging I found the issue about the missing `status.observedGeneration` field:

It boils down to the subtle differences between the claim and the composite. The composites have `xpv1.ResourceStatus  `json:",inline"` as an anonymous field in their status, where the claims don't have that (except the objectbucket for some reason).

To solve this, there are two options:
* Either we add that anonymous field to the status of all the services claims, essentially making their status implement the same interface that any composite status has to implement
* Or we use the composite to generate the XRDs, they fully implement the interface.

IMHO the second solution is the cleaner, as XRD stands for Composite Resource Definition and not Claim Resource Definition. So conceptionally using the composite to generate it makes more sense and should avoid such issues in the future.

However, the second approach could be a PITA as the whole generator logic for the XRD currently assumes that it works with a claim and injects the  `X` prefix at various places for the names.

zugao:
> However, the second approach could be a PITA as the whole generator logic for the XRD currently assumes that it works with a claim and injects the `X` prefix at various places for the names.

First, thanks for your investigation. Second, the whole approach is already a PITA as we are using the claim to generate the XRD which means we decided on a less cleaner solution long time ago. Nevertheless I tried to generate from the composite and it will be absolutely nightmare to fix everything, even the comments have to be somewhat updated as they are not good for the XRD we generate. We will need to do some research how we could improve the whole generation topic before we can have a long lasting solution. I have already implemented the first solution you proposed in the appcat PR.

Kidswiss (approved):
LGTM now.

But please run the e2e-test to see if everything is provisioning correctly before merging.

PR #691, opened 2025-04-01
Increase Crossplane poll rate

Kidswiss:
This change will decrease the load on the K8s API servers by 5-10x.

There's no impact on the time it takes until changes are reflected on the claims or composites. Crossplane will reconcile unready composites more aggressively. The 1h poll rate only applies to composites that are considdered in a ready and synced state.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/class/defaults.yml
+++ b/class/defaults.yml
@@ -148,6 +148,7 @@ parameters:
           enabled: true
         args:
           - --enable-usages
+          - --poll-interval=1h
         image:
           repository: ${appcat:images:crossplane:registry}/${appcat:images:crossplane:image}
           tag: ${appcat:images:crossplane:tag}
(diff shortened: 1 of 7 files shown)


wejdross:
this change makes a lot of sense, but even though I agree with 1 hour, I'd start with something smaller like 10mins (5 mins default, right?)

Kidswiss:
> this change makes a lot of sense, but even though I agree with 1 hour, I'd start with something smaller like 10mins (5 mins default, right?)

The default is 1 minute.

We can also start with 10 minutes. But we'd still have to come up with something to trigger the postgresql restarts if the `sgcluster` has the `pendingrestart` condition set.

Kidswiss:
Closing in favor of #695

PR #696, opened 2025-04-04
Remove Codey comp-function

Kidswiss:
Codey doesn't currently have any logic in the comp-functions, so it will fail to reconcile if this is included.

## Checklist

      changelog.
      linked in the changelog.


Kidswiss:
> THis update will leave Codey outside from our new CI/CD pipeline. Was this tested in lab? I would test with an existing instance and a new one

Well codey itself is just a PnT composition which nests a VSHNForgejo. So the relationship is similar like VSHNKeycloak and VSHNPostgreSQL. The underlying VSHNForgejo maintenance job will handle the switching of the revisions.

PR #697, opened 2025-04-08
Chose default backup region according to the OCP cluster zone

vshnbot:
## Summary

By default the component will now check what region the current OCP cluster is on. Then it will compare against a list of regions, filter out its own region and then take the first one.

It's also possible to override the region if the customer has a special requirement for that.

## Checklist

- [x] Update tests.

Link: https://github.com/vshn/appcat/pull/342


class/defaults.yml
```diff
@@ -126,6 +126,21 @@ parameters:
     grpcEndpoint: host.docker.internal:9443
     proxyFunction: false
 
+    bucketRegionOverride: ""
+
+    cloudRegionMap:
+      cloudscale:
+        - lpg
```
zugao, line 133: it's not rma1 and lpg1 right?

zugao (approved):
LGTM

PR #704, opened 2025-04-20
Fix various splitControl plane issues

TheBigLee:
## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/component/vshn_appcat_services.jsonnet
+++ b/component/vshn_appcat_services.jsonnet
@@ -30,7 +30,6 @@ local getServiceNamePlural(serviceName) =
     serviceNameLower + 's';
 
 local vshn_appcat_service(name, serviceParams) =
-  local isOpenshift = std.startsWith(inv.parameters.facts.distribution, 'openshift') || inv.parameters.facts.distribution == 'oke';
   local isBestEffort = !std.member([ 'guaranteed_availability', 'premium' ], inv.parameters.facts.service_level);
 
   local connectionSecretKeys = serviceParams.connectionSecretKeys;
@@ -112,7 +111,7 @@ local vshn_appcat_service(name, serviceParams) =
                         plans: std.toString(plans),
                         defaultPlan: serviceParams.defaultPlan,
                         quotasEnabled: std.toString(params.services.vshn.quotasEnabled),
-                        isOpenshift: std.toString(isOpenshift),
+                        isOpenshift: std.toString(vars.isServiceClusterOpenShift),
                         sliNamespace: params.slos.namespace,
                         ocpDefaultAppsDomain: params.services.vshn.ocpDefaultAppsDomain,
                         ownerKind: xrd.spec.names.kind,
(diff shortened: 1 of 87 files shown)


component/vshn_appcat_services.jsonnet
```diff
@@ -110,7 +110,7 @@ local vshn_appcat_service(name, serviceParams) =
                         plans: std.toString(plans),
                         defaultPlan: serviceParams.defaultPlan,
                         quotasEnabled: std.toString(params.services.vshn.quotasEnabled),
-                        isOpenshift: std.toString(isOpenshift),
+                        isOpenshift: if vars.isControlPlane then std.toString(params.services.vshn.isServiceClusterOpenshift) else std.toString(isOpenshift),
```
TheBigLee, line 113: I'm not really happy with this approach, if someone has a better idea, I'm all ears :-)

Kidswiss: Can't you use the `distribution` fact to make this automagic?

zugao: Maybe you could move this whole if else clause into `vars` file? Give it a nice explicit name and you are good to go

zugao: Have you talked with aldebaran? they might be able to help us. These isCluster or isControlePlaneOpenshit - all this smells very bad. It will bite us later.

TheBigLee: I've moved the whole logic to `vars`. Please check again if this solution works for you

tests/dev.yml
```diff
@@ -34,7 +34,7 @@ parameters:
         registry: dockerhub.vshn.net
 
     grpcEndpoint: 172.19.0.1:9443
-    proxyFunction: true
+    proxyFunction: false
```
zugao, line 37: I think we agreed that this must remain in true

tests/golden/dev/appcat/appcat/apiserver/30_deployment.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   annotations:
-    metadata.appcat.vshn.io/enabled-service-hash: f1a27f82c3b5113e47030d7abaa6d596
+    metadata.appcat.vshn.io/enabled-service-hash: 3265bd6918e805528a293019e7e7d136
```
zugao, line 5: @Kidswiss this seems to change even when the enabled vshn service list remains the same.... the `std.md5` is not working properly?

Kidswiss: `std.md5` should always produce the same hash for the same input. Maybe the `foldl` is missbehaving?

tests/dev.yml
TheBigLee: oh ok. I missed that. I'll revert that then

tests/golden/dev/appcat/appcat/apiserver/30_deployment.yaml
TheBigLee: This happened because I've disabled the `proxyFunction`. I've re-enabled it and the hash is now the same again

zugao (approved):
This one make sure we don't delete stuff from existing clusters!

Kidswiss (approved):

PR #740, opened 2025-06-03
Fix unescaped function names and dedup them

Kidswiss:
This fixes two issues:

- One function was generated with name that contains dots which is invalid on k8s
- With that fixed it can happen that dupes are generated. It now also filters those as well

So currently it's a bit a racecondition if argo deploys the broken one first or the right one first, which can leave to broken reconciles.

## Checklist

      changelog.
      linked in the changelog.


zugao (approved):
LGTM

PR #742, opened 2025-06-03
Upgrade crossplane dependencies

zugao:
This PR upgrades provider kubernetes to 1.18.0 and Crossplane to 1.20.

## Checklist

      changelog.
      linked in the changelog.


class/defaults.yml
```diff
@@ -172,6 +172,8 @@ parameters:
         args:
           - --enable-usages
           - --poll-interval=1h
+          - --enable-realtime-composition=false
+          - --registry=xpkg.upbound.io
```
Kidswiss, line 176: What's the rationale behind this?

Kidswiss (approved):

class/defaults.yml
Kidswiss: Ah,they changed repo again...

Kidswiss:
I tried out 1.20.0 in kind. I think they broke something :D

They removed the `/cache` mount in the chart. It now has multiple subfolders mounted under `/cache`.

However; when trying to deploy a function or provider it fails, because it still wants to write to `/cache` and not to the subfolders...

So we need to overwrite this in the helm values for the time being:
```
extraVolumeMountsCrossplane:
          - mountPath: /cache
            name: package-cache
```

I'll probably open an issue upstream later...

zugao:
> I tried out 1.20.0 in kind. I think they broke something :D
> 
> They removed the `/cache` mount in the chart. It now has multiple subfolders mounted under `/cache`.
> 
> However; when trying to deploy a function or provider it fails, because it still wants to write to `/cache` and not to the subfolders...
> 
> So we need to overwrite this in the helm values for the time being:
> 
> ```
> extraVolumeMountsCrossplane:
>           - mountPath: /cache
>             name: package-cache
> ```
> 
> I'll probably open an issue upstream later...
> 
> EDIT: okay that breaks it in a different way :D ArgoCD is now complaining about duplicate keys in the deployment

Do you recommend to not deploy for time being?

Kidswiss:
> It's strange you have such a problem in your setup. I don't see any issues with that in upstream https://github.com/crossplane/crossplane/issues?q=is%3Aissue%20state%3Aopen%20%20cache. Since you have all the details can you open the issue there?

Were you able to reproduce the problem? It's possible I did something wrong.

zugao:
> Were you able to reproduce the problem? It's possible I did something wrong.

This will have to be further tested whoever took over that ticket.

PR #801, opened 2025-07-01
Release appcat

vshnbot:
## Summary

* This is a release PR

Link: https://github.com/vshn/appcat/pull/400


Kidswiss:
This has to come from the `develop` branch. Not a separate one, otherwise we'll still have divergance between those two.

PR #825, opened 2025-07-14
Consolidate Crossplane version configuration

Kidswiss:
Only configure the Crossplane version with one single configuration.

## Checklist

      changelog.
      linked in the changelog.


class/defaults.yml
```diff
@@ -174,10 +169,6 @@ parameters:
           - --poll-interval=1h
           - --enable-realtime-compositions=false
           - --registry=xpkg.upbound.io
-        image:
-          repository: ${appcat:images:crossplane:registry}/${appcat:images:crossplane:image}
-          tag: ${appcat:images:crossplane:tag}
-          pullPolicy: ${appcat:images:crossplane:pullPolicy}
```
zugao, line 180: We might want to keep the policy? especially now that we want to change it in lab to `Always`.

PR #835, opened 2025-07-17
Improve version list generation script

Kidswiss:
This script is now written in Python. It will also ensure that it only looks at the last 5 tags from the closest tag according to the current commit.

## Checklist

      changelog.
      linked in the changelog.


Kidswiss:
> Looks fine but I would definitely add a unit test.

Not really sure how I could add a unit test to simulate a git repository though.

PR #856, opened 2025-07-30
Fix e2e connection job assertion

Kidswiss:
Somewhere between k8s 1.29 and 1.31 jobs received a new condition c`type `SuccessCriteriaMet` see https://kubernetes.io/docs/concepts/workloads/controllers/job/#success-policy.

However, this new condition confused our Kuttl tests as it expected a condition with type `Complete` as the first element of the array.

Also, the deletion protection is now only enabled during the actual test. Otherwise it's disabled, this prevents left over instances, if the tests fail for some reason before the deletion protection gets removed.

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/tests/e2e/common/scripts/check-protection.sh
+++ b/tests/e2e/common/scripts/check-protection.sh
@@ -7,12 +7,16 @@ name="$2"
 
 ns=$(kubectl -n "$NAMESPACE" get "$type" "$name" -oyaml | yq -r '.status.instanceNamespace')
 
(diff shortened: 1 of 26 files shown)


mdnix (approved):
LGTM
Did you run them again Lab?

Kidswiss:
> LGTM Did you run them against Lab?

Well, fixing this issue showed a few more :D

Here you go: https://github.com/vshn/component-appcat/actions/runs/16643212916/job/47098681490

PR #912, opened 2025-09-09
CNPG namespace: node selector change

ThisIsntTheWay:
Instead of redefining the annotation, reference it through a variable.


class/defaults.yml
```diff
@@ -206,8 +206,7 @@ parameters:
     cnpg:
       namespace: syn-cnpg-system
       namespaceLabels: {}
-      namespaceAnnotations:
-        openshift.io/node-selector: node-role.kubernetes.io/infra=
+      namespaceAnnotations: {}
```
Kidswiss, line 209: Instead of removing it:

```suggestion
      namespaceAnnotations: ${appcat:namespaceAnnotations}
```

ThisIsntTheWay:
Not sure why the golden (defaults) check is failing, doing `make golden-diff -e instance=defaults` locally works just fine...

Kidswiss:
> Not sure why the golden (defaults) check is failing, doing `make golden-diff -e instance=defaults` locally works just fine...

Can you do a `make gen-golden-all` locally again? There are some diffs in the action related to the new tag released yesterday.

mdnix (approved):

ThisIsntTheWay:
Had to rebase and _then_ the make stuff worked.

PR #920, opened 2025-09-17
Increase timeouts for e2e tests

Kidswiss:
The lab has very limited resources, but it also has autoscaling enabled so it might take a while until things get ready.

Also increasing the reliability of the backup checks for Keycloak and PostgreSQL.

https://github.com/vshn/component-appcat/actions/runs/17802518891

## Checklist

      changelog.
      linked in the changelog.


tests/e2e/forgejo/10-assert.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: kuttl.dev/v1beta1
 kind: TestAssert
-timeout: 180
+timeout: 1800
```
zugao, line 3: wouldn't this make tests non fail-able for 30 min? Is this acceptable in local env when trying to test e2e?

zugao: ah ok, haven't realised that.

zugao (approved):

PR #1133, opened 2026-04-07
Fix CNPG cluster spread alert

TheBigLee:
The alert considers every pod starting with `postgresql-`. This includes all the backup pods as well. Since the backup pods are not cleared up immediately, this alert fires all the time.
Restricting the alert to only the cluster pods, solves the issue.

## Checklist

      changelog.
      linked in the changelog.
- [ ] If no changes have been made you can self approve but otherwise at least two reviews are required

### DIFF
--- a/component/cnpg.jsonnet
+++ b/component/cnpg.jsonnet
@@ -217,7 +217,7 @@ local prometheusrule = std.prune(kube._Object('monitoring.coreos.com/v1', 'Prome
           },
           {
             alert: 'CNPGClusterZoneSpreadWarning',
-            expr: 'count by(namespace)(kube_pod_info{namespace=~"vshn-postgresql-.*", pod=~"postgresql-.*"}) > on(namespace) count by(namespace)(count by(namespace, label_topology_kubernetes_io_zone)(kube_pod_info{namespace=~"vshn-postgresql-.*", pod=~"postgresql-.*"} * on(node) group_left(label_topology_kubernetes_io_zone) kube_node_labels)) unless on(namespace) ALERTS{alertname="CNPGClusterInstancesOnSameNode", alertstate="firing"}',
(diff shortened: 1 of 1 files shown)


mikeshootzz:
@TheBigLee does this actually work? CNPG does not use an STS since the pods are directly managed by the operator

Kidswiss:
Guess I missed one alert here: https://github.com/vshn/component-appcat/pull/1101/changes

Also, as @mikeshootzz your solution might not work as CNPG doesn't use STS. My fix actually looks at the labels and filters via them.

TheBigLee:
> @TheBigLee does this actually work? CNPG does not use an STS since the pods are directly managed by the operator

You are right. The correct label is not `StatefulSet` but `Cluster`

TheBigLee:
> Guess I missed one alert here: https://github.com/vshn/component-appcat/pull/1101/changes
> 
> Also, as @mikeshootzz your solution might not work as CNPG doesn't use STS. My fix actually looks at the labels and filters via them.

Hm. Not sure If I should use your solution or stick with mine. WDYT?

mikeshootzz:
> > Guess I missed one alert here: https://github.com/vshn/component-appcat/pull/1101/changes
> > Also, as @mikeshootzz your solution might not work as CNPG doesn't use STS. My fix actually looks at the labels and filters via them.
> 
> Hm. Not sure If I should use your solution or stick with mine. WDYT?

For consistency’s sake I would stick with the solution we already have

Kidswiss:
> > Guess I missed one alert here: https://github.com/vshn/component-appcat/pull/1101/changes
> > Also, as @mikeshootzz your solution might not work as CNPG doesn't use STS. My fix actually looks at the labels and filters via them.
> 
> Hm. Not sure If I should use your solution or stick with mine. WDYT?

I'd use the same solution. Because otherwise it gets more complicated.

TheBigLee:
> > > Guess I missed one alert here: https://github.com/vshn/component-appcat/pull/1101/changes
> > > Also, as @mikeshootzz your solution might not work as CNPG doesn't use STS. My fix actually looks at the labels and filters via them.
> > 
> > 
> > Hm. Not sure If I should use your solution or stick with mine. WDYT?
> 
> I'd use the same solution. Because otherwise it gets more complicated.

Ok. Fixed to change to use the same as approach as you did earlier

zugao (approved):

mikeshootzz (approved):

Kidswiss (approved):

mdnix (approved):

PR #1174, opened 2026-05-06
Add ability to disable maintenance via component

vshnbot:
## Summary

* This is handled completely within the job itself, circumventing the release management
* A new configmap is deployed to the `appcat-control` namespace which will contain configuration for each service
* If the configmap or any service is not defined it will assume everything is enabled, otherwise take the settings from the configmap

## Checklist

- [ ] Update tests.
- [ ] Merge with `/merge` comment.

Link: https://github.com/vshn/appcat/pull/668


TheBigLee (approved):

mikeshootzz (approved):

tests/dev.yml
```diff
@@ -207,6 +207,9 @@ parameters:
 
         postgres:
           sgNamespace: stackgres
+          # maintenance:
+          #   disableServiceMaint: false
+          #   disableAppcatRelease: true
```
zugao, line 212: the config map generated has `disableAppcatRelease: true`, you might want to remove the comment here or regenerate tests

zugao (approved):
Just one small thing to look and then you can just merge

tests/dev.yml
zugao: I would use a golden test to have service maintenance disabled just fur coverage sake.

Kidswiss: good catch.

had to re-run `gen-golden-all` anyway due to the merge conflicts

PR #1210, opened 2026-06-12
Bump Crossplane to v2.2.2

mikeshootzz:
## Summary

This bumps Crossplane to v2.2.2 in the defaults, which allows us to unpin all clusters that are currently pinned to v2.2.0.
The regular monthly update PR will further update this to v2.3, but this will need more testing

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/class/defaults.yml
+++ b/class/defaults.yml
@@ -6,7 +6,7 @@ parameters:
     charts:
       crossplane:
         source: https://charts.crossplane.io/stable
-        version: 1.20.1
+        version: 2.2.0
       redis:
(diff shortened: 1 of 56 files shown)


zugao (approved):
We already have v2.2.2 which has security patches. Maybe we could do it here?

mikeshootzz:
> We already have v2.2.2 which has security patches. Maybe we could do it here?

Fair point. The idea was to have a no-op but checking the release notes this makes sense. Thanks

Kidswiss (approved):

